PR #69, opened 2014-12-29
Take into account multiple parted certs

zag2art:


zag2art:
I got that the PR is not necessary. Chained certs are required only for browsers. Here we should use only  main cert itself.

PR #70, opened 2015-01-08
add passReqToCallback parameter

zag2art:
It's usual for passport strategies - f.e. look at passport-local strategy.


ploer:
Sure.  Can you add a test & update documentation as well?  Thanks!

ploer:
Merged #71, which is near-duplicate.

PR #73, opened 2015-02-04
Add parameter @ForceAuthn to SAML AuthnRequest. 

przemko1984:
Hi,

Is it possible to add @ForceAuthn to SAML AuthnRequest? It is configurable by forceAuthn boolean option (default false)


greaterweb:
:+1: This is a feature we require for our project using passport-saml, I hope this will be accepted

ploer:
Thanks for the PR, we can definitely add this.

I think I'd prefer if it only emited the ForceAuthn field if true, though.  (in the interests of just not changing existing behaviors if not neccessary).  Can you update the PR with that change?

Could you also include a proposed change to the readme to document the new option?

przemko1984:
Yep, I updated it.

ploer:
Merged & published as 0.9.0.

PR #75, opened 2015-02-05
adds error handling for missing user in logout routine

jlippold:
I had some trouble getting log out to work, so I added some error handling with a callback when no user is defined


jlippold:
I see the tests are failing, i'll give it another look in a bit

PR #80, opened 2015-03-04
Pass 'RelayState' from query string into req body

frumioj:
I didn't find an easy way to "automatically" add a dynamic RelayState to the SAML AuthN request, so I made this change to the passport-saml strategy. If you put a RelayState query parameter on the request that calls the passport-saml middleware, the req.body.RelayState is set to this value. This allows you to, for example, put the actual request URL (encoded appropriately) into the RelayState value (which used to be a common thing to do in SAML implementations). I expect you can probably implement a passport cache provider for this, but this seemed simpler and is effective enough (for now).

### DIFF
--- a/lib/passport-saml/strategy.js
+++ b/lib/passport-saml/strategy.js
@@ -28,6 +28,10 @@ Strategy.prototype.authenticate = function (req, options) {
 
   options.samlFallback = options.samlFallback || 'login-request';
 
+  if ( req.query.RelayState ) {
+      req.body.RelayState = req.query.RelayState ;
+  }
+  
   function validateCallback(err, profile, loggedOut) {
       if (err) {
         return self.error(err);


ploer:
My apologies, I just realized I somehow missed this PR when you submitted it.

I'm afraid I don't understand the problem this is solving, though.  If you have a req.query.RelayState set, passport-saml already picks it up, right?  (saml.js:261).   What is the benefit of copying it over into req.body.RelayState?

beeva-pabloterrado:
Hello both,

I have developed a portal in nodeJs and angularJs and I am actually using passport-saml v0.9.2.
My server can scale (increase number machines) and may be the server who makes the request dont receive the SAML response if there is a lot of requests.

I have two important questions:

1) How I can implement a cacheProvider object and use it instead the inmemory-cache-provider who is being used actually?

2) Actually, user goes to my url, and is redirected to /login which redirects to the login url of the IdP.
But if I want to make a link to a private url content of my web from other url.
For example: from www.otherdomain.com introduce a link to https://mydomain.com/content/yyy
And in my file routes.js, I get that route and I can pass "yyy" to auth/saml/login?redir=yyy.

And in router.get('/login,
function(req,res,next){
    if(req.query.redir){
       var customConfig = _.cloneDeep(config);
       customConfig.saml.aPRS.RelayState='redir'+req.query.redir;
       require('./passport').setup(User, customConfig);
    }
    else{
       require('./passport').setup(User, config);
    }
   passport.authenticate('saml')(req,res,next);
}

But if I do that, I lost my cache provider. ¿How can I do for change dinamically RelayState in each request without losing this inmemory cache provider implemented by passport-saml?

Some idea?

Thank you so much.

CatBakun:
```javascript

// supose your app gets called like this: /some/private/content?some_param=HelloAuthenticatedWorld

app.get('/some/private/content/', 
  function thisWillDoTheTrick(req, res, next) {
    if(req.session.user){
     next();
     return;
    }
        'RelayState': '/some/private/content?bypassed_value=' + req.query.some_param
    })(req, res, next) // <-- note that here we are executing passport.authenticate.
  },
  function hereIAmLoggedIn(req, res, next) {
    res.send(req.query.bypased_value); // HelloAuthenticatedWorld
  })
```

PR #87, opened 2015-04-23
Add Destination attribute to LogoutResponse

cmordue:
Add Destination attribute to the LogoutResponse to be compliant with the spec.

See section 3.4.5.2 and 3.5.5.2 of http://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf
> If the message is signed, the Destination XML attribute in the root SAML element of the protocol message MUST contain the URL to which the sender has instructed the user agent to deliver the message. The recipient MUST then verify that the value matches the location at which the message has been received.

Without the Destination attribute, PingFederate (at least in 6.10) will reject the LogoutResponse.


ploer:
Thanks for the pull request!  Merged & published as 0.9.2.

PR #89, opened 2015-05-04
Fix default is https

vlkous:
Does not fit with documentation


ploer:
Thanks for submitting this, and sorry for the slow response!

Instead of changing the default I will change the documentation, though.

ploer:
Rationale is that this would be a breaking change in behavior.

PR #92, opened 2015-05-22
Fix issue where query strings in entry points mess up entryPoint URLs

brianhartsock:
Noticed an issue where you a query string in an entrypoint will not be handled correctly. For example:

```
{
   entryPoint: 'https://exampleidp.com/path?key=value'
}
```
would end up causing a URL such as `https://exampleidp.com/path?key=value?SAMLRequest=...`

The existing tests were a little complicated for me to figure out where to throw some tests to validate behavior. I decided to write some simple tests separately to expose the issue and validate this fixes it.

Let me know if there is anything else you would like to see.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -3,6 +3,7 @@ var xml2js = require('xml2js');
 var xmlCrypto = require('xml-crypto');
 var crypto = require('crypto');
 var xmldom = require('xmldom');
+var url = require('url');
 var querystring = require('querystring');
 var xmlbuilder = require('xmlbuilder');
 var xmlenc = require('xml-encryption');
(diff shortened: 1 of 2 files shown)


ploer:
Hmm, never thought of the use case of having an entrypoint with a query parameter already in it.  Thanks for the PR!

My once initial concern is that this change still uses querystring to generate the signature around line 250.  It probably works, but it seems brittle to generate the signature with one codepath, but then actually use the output of a differnet codepath.

Can you update that to use url as well?  (and just strike the require for querystring)

brianhartsock:
`url` actually uses `querystring` [under the hood](https://nodejs.org/api/url.html#url_url_parse_urlstr_parsequerystring_slashesdenotehost) and can't format just the query string. Since you are only signing the SAML message portion of the query string and not the full URL, I think switching to sign the entire URL would break things.

I could be wrong, I haven't dug into the signing logic too much yet. Thoughts?

brianhartsock:
@ploer Following up on this thread. Thoughts on my last comment?

ploer:
Okay, I think that makes sense.  Will merge.

Issue #81, opened 2015-03-16
Make passport-saml work with wso2 is,


Indrani123:
I am using passport-saml nodejs module with WSO2is-5.0.0 using SAMLstragegy  trying to implementt SSO, I am able to login WSO2is-5.0.0 server, using it, but when I am try to logout , my session still exist, unable to implement logout .

here my SAMLStrategy configuration:

    passport.use(new SamlStrategy(
    {
        path: '/AssertionConsumer',
        entryPoint: config.idpUrl,
        issuer: config.issuer,
        protocol: config.httpProtocol,
        identifierFormat : config.identifierFormat,
        logoutUrl:'/logout',
        attributeConsumingServiceIndex:1012175983,
       }, function(profile, done){

        console.log('Profile: %j', profile);
        return done(null, profile);
    }
      ));

Is there any configuration I miss for wso2is-5.0.0 with passport-saml ?

This is my config data:

    config.idpUrl= "https://localhost:9443/samlsso";
    config.issuer = "passport-saml";
    config.httpProtocol = "http://";
    config.identifierFormat =  "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress";

I had checked further, starting wso2 in debug mode

log4j.logger.org.wso2.carbon.identity=DEBUG

then  using a logout url in route also configure in strategy

   logoutUrl:'/logout'

     router.get('/logout',passport.authenticate('saml', { failureRedirect:  '/error', failureFlash: true, samlFallback:'logout-request' }),function(req, res) {
   
    req.logout();
    res.redirect('/');

     });

now when I click further in http://localhost:9001/logout

I got this error : Error when processing the authentication request! msg in browser

so I did check in my logs found passport-saml have missing SessionIndex information in LogoutRequest

        <?xml version="1.0"?><samlp:LogoutRequest xmlns:samlp="u
          rn:oasis:names:tc:SAML:2.0:protocol"  xmlns:saml="urn:oasis:names:tc:SAML:2.0:ass
       ertion" ID="_66a8e1407de73b21ed44" Version="2.0" IssueInstant="2015-03-16T16:06:
       41.743Z" Destination="https://localhost:9443/samlsso"><saml:Issuer  xmlns:saml="u
          rn:oasis:names:tc:SAML:2.0:assertion">passport-saml</saml:Issuer><saml:NameID Fo
          rmat="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress">admin</saml:NameID
[...]

PR #93, opened 2015-05-26
SessionIndex / WSO2 logout

vesse:
This PR adds `sessionIndex` property to returned profile if one is defined in `AuthnStatement`. In addition it adds `saml2p:SessionIndex` element to the generated logout request if `req.user.sessionIndex` exists.

Related to #81. Tested with WSO2 5.0.0 with following logout URI handler:

```javascript
app.get('/logout', function(req, res) {
  if (req.user == null) {
    return res.redirect('/');
  }
  return samlStrategy.logout(req, function(err, uri) {
    return res.redirect(uri);
  });
});
```

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -186,6 +186,36 @@ SAML.prototype.generateLogoutRequest = function (req) {
   return xmlbuilder.create(request).end();
 };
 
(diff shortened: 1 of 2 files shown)


ploer:
Thanks for the PR, will merge.

Is there any update to the documentation page that would be useful for users who need this functionality?  If so, a PR to add that would be great.

PR #101, opened 2015-07-31

achingbrain:
We're using passport-saml in production and it's great (thanks!).  We've got ADFS automatically updating the service provider metadata, but the current implementation doesn't provide a SingleLogoutService callback URL.

I've added this information to the generated metadata and added a test for it too.  We've been happily running this in production against ADFS 3.0 for several weeks without issue.

I found that the order of elements in the metadata file is important, which is why I removed the inline addition of the `NameIDFormat` and `AssertionConsumerService` elements and instead added them after adding the `SingleLogoutService` bindings (if a `logoutCallbackUrl` option is supplied).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules/
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Config parameter details:
 * Logout
  * `logoutUrl`: base address to call with logout requests (default: `entryPoint`)
  * `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
+ * `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
 
 ### Provide the authentication callback
 
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -784,18 +784,34 @@ SAML.prototype.generateServiceProviderMetadata = function( decryptionCert ) {
       '@xmlns': 'urn:oasis:names:tc:SAML:2.0:metadata',
       '@xmlns:ds': 'http://www.w3.org/2000/09/xmldsig#',
       '@entityID': this.options.issuer,
+      '@ID': this.options.issuer.replace(/\W/g, '_'),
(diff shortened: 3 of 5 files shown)


doodlemoonch:
+1

ploer:
Thanks for the PR, @achingbrain.  Generally looks good to me but two requests:

(1) Can you explain to me why you pulled the  NameIDFormat and AssertionConsumerService down to line 809 in saml.js?  Since they are still unconditionally inserted, at first glance it's not clear to me what that changes.

(2) Can you update your test to do it's own full expected metadata comparison?  Right now it looks like it just checks that a "SingleLogoutService" declaration exists, but unless I"m missing something I don't think it's checking the specific bindings and location output.  Would be nice to get this under test to catch if it is accidentally perturbed in any future change.

Thanks!

ploer:
Also, I just merged PR #102, which is going to create some conflicts here, sorry about that. :-)

ploer:
Oh, I read your comments again and realized you addressed my question #1.  But I'm a little worried by the implications there -- any idea why order of elements would matter?

cjbarth:
Is there anything that I can help with to get this PR completed?

ploer:
@cjbarth , @achingbrain , honestly the only reason I haven't merged this is I haven't gotten around to working through merge conflicts.  I still intend to get to it... but if someone wanted to resubmit this with a clean merge, that would probalby help.  Sorry for the delays here!

cjbarth:
@ploer , since I don't have rights to resolve the merge conflicts, I'll create a new PR with the changes merged in...unless you know of a better way.

achingbrain:
I no longer have access to the systems this patch was for - otherwise I'd rebase the change and resolve the merge conflicts, sorry.

cjbarth:
I've created a new pull request that should include all the same changes that PR #101 originally had. If there are any issues, please let me know.
Chris

Date: Fri, 16 Oct 2015 06:50:03 -0700

cjbarth:
Bergie,

 

I’ve created a new PR for you for the logout matter. Please let me know if you have any issues with the new PR.

 

Thanks,

Chris

ploer:
Thanks, will merge from the other PR.

PR #103, opened 2015-07-31
generateServiceProviderMetadata: remove callbackUrl dependency

gnawhleinad:
This change mirrors `generateAuthorizeRequest`'s behavior of constructing `callbackUrl` from `protocol` and `path` if `callbackUrl` is not supplied.

I kept `req` independent from `generateServiceProviderMetaData` by adding a `host` option that defaults to `localhost`.


ploer:
Looks good to me, can I get you to resolve the conflicts I just created, and maybe add a test case under generateServiceProviderMetadata that constructs the callback url from protocol/path/host?

Thanks!

gnawhleinad:
Done!

Issue #94, opened 2015-06-11
Signature namespace defined in parent


darrenmce:
I'm having an issue where in an Assertion, the namespace of the signature is defined in the Response element. 

```xml
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" 
xmlns:dsig="http://www.w3.org/2000/09/xmldsig#">
...

    <dsig:Signature>      
...
    </dsig:Signature>
```

So when the validateSignature method extracts the node and converts it to a string, the namespace is lost (and xml-crypto rejects since it queries based on the namespace-uri)

Is this something passport-saml cares about? or is this a bad use of the Signature element/namespace?

Temporarily, I have put a workaround in the validateSignature function for myself:

> Instead of simply `var signature = signatures[0].toString();` in [line 347](https://github.com/bergie/passport-saml/blob/8e84455a530e1694c1fe01bd3e7647128b0f1885/lib/passport-saml/saml.js#L347)

```js
  var signature = signatures[0];

  // If Signature doesn't have it's namespace explicitly defined (was defined in parent)
  // it will be lost in the toString(), in this case, add it
  var prefix = signature.prefix;
  var nsAttr = 'xmlns';
  if (prefix) {
    nsAttr += ':' + prefix;
  }
  if (!signature.hasAttribute(nsAttr)) {
    signature.setAttribute(nsAttr, 'http://www.w3.org/2000/09/xmldsig#');
  }

  signature = signature.toString();
```

ploer:
Hmm, interesting.  I'll take a look at this, but might be a while before I get around to it (sounds like your workaround is sufficient for the moment).

brianhartsock:
@darrenmce Had the same issue. See https://github.com/yaronn/xml-crypto/pull/59 where I'm trying to get a few changes into xml-crypto. Discussion is ongoing.

darrenmce:
@brianhartsock ah, great to know its not just me. I took a look at that issue in xml-crypto, and though its not clear to me where the fix belongs its becoming clear to me that a fix is necessary:

response from Oracle:

---------------------- 

OAM 11gR2 federation declares the Signature in the ancestor element of the Signature element. 

example: 

```xml
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" 
xmlns:dsig="http://www.w3.org/2000/09/xmldsig#" 
xmlns:enc="http://www.w3.org/2001/04/xmlenc#" 
[...]

PR #104, opened 2015-08-04
Fixed an obscure bug in which certificates may not be found.

ethanmick:
I did not write up a test case, but I did have to change this in my own implementation,
and it fixed the bug. If you'd like, I will attempt to write an actual test case.

This also more closely mimics the tests in xml-crypt, as [shown here](https://github.com/yaronn/xml-crypto/blob/master/test/saml-response-test.js#L8).

Not sure if anyone else has run into this issue, but I did in a separate project.
The issue is that when the signature node has been serialized to a String,
it loses the context of where it is in the XML tree. When deserialized, it
only knows about itself. Later in the code, in valdiateSignatureValue of xml-crypto,
this is called:

`var signedInfo = utils.findChilds(this.signatureNode, "SignedInfo")`

However, if the signature node has been serialized and deserialized, it
incorrectly reports it only has 1 child (itself), as opposed to the correct number.
This causes the certificate data to not be found, and the validation to fail.

All tests still pass with this change. Update to xml-crypto required.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -344,7 +344,7 @@ SAML.prototype.validateSignature = function (fullXml, currentNode, cert) {
   //   than that, reject.
   if (signatures.length != 1)
     return false;
-  var signature = signatures[0].toString();
+  var signature = signatures[0];
   var sig = new xmlCrypto.SignedXml();
   sig.keyInfoProvider = {
     getKeyInfo: function (key) {
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "passport": "0.2.x",
     "q": "1.1.x",
     "xml2js": "0.4.x",
-    "xml-crypto": "0.3.x",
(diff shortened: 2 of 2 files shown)


ploer:
Nice catch, thanks.  If you were able to construct a test case that exposes the bug and shows it fixed, that would be awesome.

I should be able to take a closer look and merge a little later this week.

ploer:
Ah, looks like this is picking up the change from https://github.com/yaronn/xml-crypto/pull/59, discussed in issue #94.  Somehow I'd missed that that had gotten merged in xmlcrypto.

brianhartsock:
@ploer awesome. how frequently do you cut additional NPM releases? Just wondering when to check back so I can use it vs. a git hash to depend on.

ploer:
Was hoping to include PR #105, but looks like that's not quite ready, so will go ahead with a push tonight of what's in.

brianhartsock:
@ploer great, thanks.

PR #105, opened 2015-08-05
Allow multiple AuthnContextClassRef to be configured

david-martin:
Delimited by a semi-colon `;`

For example: `urn:federation:authentication:windows;urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport`

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -128,13 +128,22 @@ SAML.prototype.generateAuthorizeRequest = function (req, isPassive, callback) {
     }
 
     if (!self.options.disableRequestedAuthnContext) {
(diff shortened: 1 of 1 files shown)


ploer:
Thanks for the PR!

A couple thoughts:

1) Can you add a test case for this?
2) Can you update the documentation to reflect the added behaviour?
3) The semicolon-delimination feels a little awkward to me... are URIs forbidden from having semicolons in them in the schema?  Would it make sense to just allow `authnContext` to optionally be an array?

ploer:
Still happy to include this with those changes if you get around to it, but will close for now due to inactivity.

PR #106, opened 2015-08-13
Fixed broken validatePostRequest method; Added support for parsing se…

brianhartsock:
…ssion index out of logout request

This PR adds support for the optional `SessionIndex` attribute of the LogoutRequest. It also fixed the broken `validatePostRequest` method which would never accurately validate a certificate due to a changed method signature.

Also added some public/private keys to the repo to help anyone that needs to sign static xml docs for testing.


brianhartsock:
@ploer any thoughts on this?

ploer:
Looks good to me, thanks!

ploer:
Published as 0.11.1.

PR #107, opened 2015-08-14
Support SHA256 signatures

timbrody:
Added support for specifying SHA256 signatures, which are the default in AD FS.
Generally SHA1 is being retired due to its cryptographic weakness.

Left the default signature as "SHA1" to maintain backwards compatibility.


ploer:
Thanks for the PR!  Two requests:

1) Can you add a test of generating a SHA256 signature?
2) Can you revise the comment in the documentation to clarify that it is setting the signature to be used in signing?  (I feel like right now it might be ambiguous on a quick read whether it in any way controls signatures being verified)

Also, if you felt inspired to refactor `signRequest()` so that the logic to populate the sigAlg field and the logic to actually generate the signature lived next to each other, that would be awesome.  (but if not, no worries)

timbrody:
I've revised the note and added a recipe for AD FS.
I've refactored signRequest() although that makes it side-affecting?

Can you give me a pointer on adding a test case? What's needed to verify the response?

timbrody:
I've added unit tests for sha256 and sha1.

ploer:
Thanks!

PR #110, opened 2015-09-07
Updated passport package dependency

dbck:
jaredhanson/passport#400

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "main": "./lib/passport-saml",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


ploer:
Thanks for the PR.  Based on reading that thread though, it sounds like the real recommendation is actually to depend on passport-strategy instead of passport directly.  I'll make that change.  Let me know if I'm missing something!

dbck:
You're right. Changing the dependency to passport-strategy will be the better solution in the long term.

PR #111, opened 2015-09-07
Ignore attribute with no attributeValue child

T1B0:
A childless attribute (without attributeValue child node) like :
```
  <AttributeStatement>
    <Attribute Name="fubar.roles"/>
  ....
 </AttributeStatement>
```
would cause a typeError in saml.js#L582. this fix would ignore such attribute

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -671,6 +671,10 @@ SAML.prototype.processValidlySignedAssertion = function(xml, inResponseTo, callb
 
       if (attributes) {
         attributes.forEach(function (attribute) {
+         if(!attribute.hasOwnProperty('AttributeValue')) {
+            // if attributes has no AttributeValue child, continue
+            return;
+          }
           var value = attribute.AttributeValue;
           if (value.length === 1) {
             profile[attribute.$.Name] = attrValueMapper(value[0]);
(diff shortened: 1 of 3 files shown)


T1B0:
Also please note that this is a quick-fix, according to SAML specs the attributeValue is defined as 

```
<element name="Attribute" type="saml:AttributeType"/>
<complexType name="AttributeType">
  <complexContent>
    <extension base="saml:AttributeDesignatorType">
      <sequence>
        <element ref="saml:AttributeValue" maxOccurs="unbounded"/>
      </sequence>
    </extension>
  </complexContent>
</complexType>
```

and since the element is inside a "sequence", the minOccurs default value is 1, so the attributeValue is suposed to appears at least once for any attribute tag. 
I don't know if you'd rather throw an error than just ignore a malformed attribute, i guess it depends on how strict you intend this lib to be.

ploer:
Thanks for the PR!  If it's something you've actually run into from a real IDP, then I'd definitely like to tolerate it.  

Can you add a test case for this?

T1B0:
just added a test case. and yes I actually run into this from a customer's private IDP, I've asked on which SAML lib they're based but didn't get an answer yet - i'll let you know if I find out

T1B0:
Just heard back from this customer, they uses a commercial provider called "sign&go" by ilex => http://www.ilex-international.com/en/our-products/signgo/ . so this is a legit, live use case I guess.

T1B0:
had some feedback from an ilex developer, he pointed out that in SAML2.0 the attributeValue has become optional : 

2.7.3.1 Element <Attribute>
The <Attribute> element identifies an attribute by name and *optionally includes its value(s)*. It has the
AttributeType complex type. It is used within an attribute statement to express particular attributes ...

the xml-schema I quoted earlier was from SAML1.1 specs, where attributeValue were to occurs 1 or more per attribute

ploer:
Thanks!

PR #113, opened 2015-09-24
Process all attribute statements instead of only first

vsimonian:
Given the following SAML assertion

```xml
<saml2:Assertion xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" Version="2.0" ID="_a293b185-490e-48ba-a30d-9ff96e9c5b2d" IssueInstant="2015-09-24T20:59:14Z">
  <saml2:Issuer>http://stubidp.kentor.se/Metadata</saml2:Issuer>
  <saml2:Subject>
    <saml2:NameID>JohnDoe@example.com</saml2:NameID>
    <saml2:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"/>
  </saml2:Subject>
  <saml2:Conditions NotOnOrAfter="2015-09-24T21:01:14Z"/>
  <saml2:AttributeStatement>
    <saml2:Attribute Name="email">
      <saml2:AttributeValue>JohnDoe@example.com</saml2:AttributeValue>
    </saml2:Attribute>
  </saml2:AttributeStatement>
  <saml2:AttributeStatement>
    <saml2:Attribute Name="uid">
      <saml2:AttributeValue>12345</saml2:AttributeValue>
    </saml2:Attribute>
  </saml2:AttributeStatement>
</saml2:Assertion>
```

the only processed attribute under the current code would be `email`.

With this new code, `uid` would be processed and made available on `profile` as well.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -663,7 +663,11 @@ SAML.prototype.processValidlySignedAssertion = function(xml, inResponseTo, callb
 
     var attributeStatement = assertion.AttributeStatement;
     if (attributeStatement) {
-      var attributes = attributeStatement[0].Attribute;
+      var attributes = [].concat.apply([], attributeStatement.filter(function (attr) {
+        return Array.isArray(attr.Attribute);
+      }).map(function (attr) {
+        return attr.Attribute;
+      }));
 
       var attrValueMapper = function(value) {
         return typeof value === 'string' ? value : value._;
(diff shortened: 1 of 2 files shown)


gnawhleinad:
We should add a test case to verify this change!

vsimonian:
@gnawhleinad Just added a test case.

ploer:
Looks good, thanks!

PR #119, opened 2015-10-16

cjbarth:


cjbarth:
This is a PR that mirrors the changes made by @achingbrain. It has been modified only where necessary to fit on master.

PR #121, opened 2015-10-17
Document the skipRequestCompression

jonasfj:
Some providers don't support SAML compression, or they don't support compressed SAML requests if the assertions is configured to be uncompressed.

So having the skipRequestCompression is nice, and highlighting it in documentation makes it easier for people configuring an SP app through trail and error (not all IDPs gives much in terms of error codes and error messages).

----
I ran into this issue when setting up and app to work with okta. Except for the SAML request compression issue this library works great (granted I had to contact okta support got even figure out that request compression was an issue, but that's their fault for not having error codes, system logs or debug information of any kind).


ploer:
Thanks for the PR!

fcorneli:
The big problem is that passport-saml does not indicate that the request has been compressed.
It should also pass a parameter `SAMLEncoding` with value:
```
urn:oasis:names:tc:SAML:2.0:bindings:URL-Encoding:DEFLATE
```
to indicate towards the identity provider that the request should be decompressed first.

PR #122, opened 2015-10-19

cjbarth:
This is a PR that mirrors the changes made by @achingbrain. It has been modified only where necessary to fit on master.


ploer:
Thanks!

PR #123, opened 2015-10-19
Do not sign custom query string parameters

cjbarth:
When using a custom query string parameter ADFS requires that only the SAMLRequest, RelayState, and SigAlg parameters be signed. Thus, this code was modified to not sign any custom query string parameters.


ploer:
Thanks for the PR.  Can you add a test case for this?

cjbarth:
Ok, I think the tests that I've modified should cover the changes I've made.

ploer:
Thanks!

PR #124, opened 2015-10-21
Add NameQualifier and SPNameQualifier to nameID

ashimaathri:
The nameID element of the authn response can contain the NameQualifier and
SPNameQualifier attribute optionally. 
[pysaml2](https://github.com/rohe/pysaml2) however, which we use to implement our identity provider, depends on these attributes being present on the logout request. So can we please add this to the logout request?


ploer:
Thanks for the pull request, looks good to me.  Can you please add a test for this new behavior, though?

ashimaathri:
Thanks! I'll add the test.

ashimaathri:
@ploer I've added the test, let me know if it needs some work.

Issue #90, opened 2015-05-06
HTTP POST Binding in passport-saml


whirp:
Hi, 
I've to use  passport-saml in  HTTP POST Binding. Would you please help me to find a way to add this implementation to this module? Is there a pre-existent module doing this?

Thanks

ploer:
You want to make requests to the IdP using POST bindings?  (responses to passport-saml are already supported via POST binding).

I'm not aware of any module that adds this functionality.  I'd be happy to consider a PR if you wanted to try to add a configuration option to support this.  (I haven't studied what this would take but you'd probably have to abstract out a bunch of hardwired assumptions in the current module).

cheton:
Hi @ploer 

I created a pull request #129 to add HTTP-POST binding support for the SAML &lt;AuthnRequest&gt;.

Since It is necessary to use HTTP-POST binding for all requests to our IdP servers, may I know when will this PR be added to passport-saml?

Thanks

PR #129, opened 2015-12-02
Adds HTTP-POST binding support for the SAML <AuthnRequest>

cheton:
See also: [HTTP POST Binding in passport-saml #90](https://github.com/bergie/passport-saml/issues/90)

In some environments, the IdP requires HTTP-POST binding for all requests including the SAML  &lt;AuthnRequest&gt;.

In this scenario, the SP should send an HTML form back to the browser in the HTTP response. The HTML form contains a SAML &lt;AuthnRequest&gt; message encoded as the value of a hidden form control named SAMLRequest. For example:

```html
<html>
<body onload="document.forms[0].submit()">
<form method="post" action="https://idp.example.com/simplesaml/saml2/idp/SSOService.php">
  <input type="hidden" name="SAMLRequest" value="request" />
  <!-- additionalParams goes from here -->
  <input type="hidden" name="RelayState" value="token" />
  <input type="submit" value="Submit" />
</form>
<script>document.forms[0].style.display="none";</script>
</body>
</html>
```

The browser will issue an HTTP POST request to the IdP's SSO service by triggering the auto submit form.

A `samlBinding` option is provided to `passport.authenticate()` by specifying `HTTP-Redirect` to enable HTTP Redirect Binding or `HTTP-POST` to enable HTTP POST Binding. For example:
```js
    samlFallback: 'login-request', // Defaults to 'login-request' if not specified
    samlBinding: 'HTTP-POST' // Defaults to 'HTTP-Redirect' if not specified
});
````

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -357,6 +357,80 @@ SAML.prototype.getAuthorizeUrl = function (req, callback) {
   });
 };
 
+SAML.prototype.getAuthorizeForm = function (req, callback) {
+  var self = this;
+
+  // The quoteattr() function is used in a context, where the result will not be evaluated by javascript
+  // but must be interpreted by an XML or HTML parser, and it must absolutely avoid breaking the syntax
+  // of an element attribute.
+  var quoteattr = function(s, preserveCR) {
+    preserveCR = preserveCR ? '&#13;' : '\n';
+    return ('' + s) // Forces the conversion to string.
+      .replace(/&/g, '&amp;') // This MUST be the 1st replacement.
+      .replace(/'/g, '&apos;') // The 4 other predefined entities, required.
+      .replace(/"/g, '&quot;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;')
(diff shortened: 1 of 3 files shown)


ploer:
I've merged this, and in penance for being so slow to get to this I went ahead and wrote the test myself.  :-)

I also changed the parameter to be passed on the base config object handed to SAML, rather than as an option to the authorize() call; let me know if you see a problem with that.

cheton:
It looks great! I really appreciate you writing the test, and I did learn a lot from it. :)

jylauril:
How exactly does this differ from the already existing `authnRequestBinding` option? I have been using that for HTTP-Post endpoints and it works just fine.. does this add something that the `authnRequestBinding` didn't provide yet?

cheton:
Hi @jylauril,

@ploer had renamed the `samlBinding` option to SAML object options `authnRequestBinding` with this commit (https://github.com/bergie/passport-saml/commit/dd32a45b5d4e3984784276dbea67f5a9e9d86db3), and HTTP-POST binding for SAML AuthnRequest support was initially added with this PR.

PR #131, opened 2015-12-11
Fix Subject dereference bug

timoruppell:
The `if (subject)` block is extended to cover all access to `subject`.

We ran into this bug in the code because our SAML provider didn't include Subject at the root of the Assertion -> line 614 throws a 'cannot access 0 of undefined' error.


ploer:
Thanks for the PR; merged!

(note that Travis tests were failing due to some scope issues; should be cleaned up now)

PR #138, opened 2016-01-25
Check for audience

sonnymai:
There doesn't seem to be any checks for the audience attribute. This would allow users using the same IDP to authenticate into service provider B if they had a valid token for service provider A (assuming both service providers use the same IDP).

Hopefully this logic is correct, I can add tests if this is verified.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -759,6 +759,15 @@ SAML.prototype.processValidlySignedAssertion = function(xml, inResponseTo, callb
         throw conErr;
     }
 
+    var audienceRestriction = conditions.AudienceRestriction ? conditions.AudienceRestriction[0] : null;
+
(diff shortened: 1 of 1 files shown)


pdspicer:
You are correct, this check is currently missing. I would contest based on the SAML specification that the audience condition validation should REQUIRE the presence of an audience element, and then further check that it contains the SP entityID.

markstos:
This feature is still of interest if someone wants to pick it up.

benjamine:
I can send a version with the suggestion above, I'm interested in getting this upstream.

benjamine:
@markstos I opened #253 picking this up, made AudienceRestriction required when options.audience is specified, and added unit tests.

markstos:
Closing this, as PR #253 was merged. Leave a comment there if you find any problem with it.

PR #141, opened 2016-02-10
Add missing arguments

XavierBoubert:
In my organisation, they use ResponseLocation in the SingleLogoutService to redirect users.
"AuthnRequestsSigned" and "WantAssertionsSigned" flags are used in the signin.

All of these params seems to be "standard" in Oracle configurations that I can found in my organisation.


XavierBoubert:
Tests fixed

pdspicer:
According to SAML metadata spec, ResponseLocation should be used if the location for sending responses is to be different than the one to which requests are sent. If they have the same value it is redundant. Suggest making this a different option like `logoutResponseCallbackUrl`

markstos:
Noting that #237 is a related PR: "added getLogoutResponseUrl"

PR #144, opened 2016-02-24
[+] new option to enable custom logout

molipet:
A new `beforeRedirect` option which can be used to implement **custom logout-logic** before _passport-saml_ redirects back to the Identity Provider.
It can be useful when the logout-logic is more complex than a simple `req.logout()`, e.g. database-pools needs to be finished properly.

In case of **Service Provider initiated logout**, it can be used with the following syntax:

```
app.get('/saml-logout',
           beforeRedirect: function(req, next) {
                  // custom logout-logic can be here
                  return next();
           },
           samlFallback: 'logout-request',
           failureRedirect: saml-logout/error',
           failureFlash: true
      })
);
```

The custom logout-logic will be executed right before _SAML LogoutRequest_ is sent to the Identity Provider.

In case of **Identity Provider initiated logout**, it can be used with the following syntax:

```
app.post('/saml-logout/callback',
           beforeRedirect: function(req, next) {
                  // custom logout-logic can be here
                  return next();
           },
           failureRedirect: saml-logout/error',
           failureFlash: true
      })
);
```

The custom logout-logic will be executed after the validation of the _SAML LogoutRequest_ but right before _SAML LogoutResponse_ is sent to the Identity Provider.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -686,14 +686,20 @@ SAML.prototype.processValidlySignedAssertion = function(xml, inResponseTo, callb
         }
       }
 
-      subjectConfirmation = subject[0].SubjectConfirmation ?
-                            subject[0].SubjectConfirmation[0] : null;
+     subjectConfirmation = null;
+
+      if (subject[0].SubjectConfirmation) {
+        for (var i = 0; i < subject[0].SubjectConfirmation.length; i++) {
+          var sc = subject[0].SubjectConfirmation[i];
+          if (sc.$ && sc.$.Method === 'urn:oasis:names:tc:SAML:2.0:cm:bearer') {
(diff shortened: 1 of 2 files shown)


pdspicer:
Assuming you have access to the SamlStrategy instance (lets call it `samlStrategy`) you are using with passport, you can instead call `samlStrategy.logout(req, callback)`, which will call back with a logout URL that you can use for redirection after executing your own custom logic. Does this satisfy your use case?

markstos:
Since there's no further reply from the OP or anyone else, I'm closing this. Note that #237 and #141 appear related. There may be some merit in the code here, but to move it forward,  someone needs to explain how it's better than what @pdspicer proposed and the other logout-related options proposed in the related PRs.

Issue #128, opened 2015-11-24
Redirects in infinite loop if a body parser is not set up for express


jrop:
Minor, but should be documented: when defining the callback route:

```
let app = express()
app.get('/login', passport.authenticate('saml', { failureRedirect: '/' }), (req, res) => res.redirect('/'))
app.post('/callback', passport.authenticate('saml', { failureRedirect: '/' }), (req, res) => res.redirect('/'))
```

There needs to be middleware that supports body parsing, otherwise (in my case) it starts redirecting between my IdP and my app indefinitely.

ploer:
Can you submit a PR with the proposed documentation change?  Thanks!

joelwass:
I seem to be having this issue, but i'm using the node package body-parser, what body parser did you end up using to resolve this issue?

casey-billett:
I feel like adding the body-parser middleware has caused functionality in Rocky to break.  When trying to reverse-proxy connections after ensuring authentication, the middleware fails with:

PhilippSpo:
I also had the same issue. Adding the `bodyParser` finally resolved it. 😅

```js
const bodyParser = require('body-parser');
app.use(bodyParser.urlencoded({ extended: false }));
```

mrchief:
Wasted days chasing this infinite loop issue. Many thanks to OP for creating this issue. There is no mention of it in documentation. :disappointed:

jrop:
@mrchief I created a PR for this a long time ago, but it's had very little activity :( #151

salazr:
Spent a couple hours this morning setting this up on a new app, and encountered the same problem.  It would be nice to have this in the documentation.

mrchief:
Its kinda in the documentation, in its own convoluted way: :wink: 

![image](https://user-images.githubusercontent.com/781818/29191638-6d0bb33a-7dec-11e7-8d69-4da08897723e.png)

MardariG:
This is closed, but I have spend too much time to find out the problem is in body-parser.
This should be as mandatory requirement on the mine README in **bold**. I don't get it why is so hard to mention this. Thanks for awesome library anyway.

DORRITO:
Hello!  If anyone is using express 5, since body parser isn't needed, simply make sure the `app.use(express.urlencoded({ extended: false }));` **_before_** the route calls.

Many thanks to kashiif and makstos above for leading me down that path.  Immense Thanks!

markstos:
[...]

PR #151, opened 2016-04-15
Add error for POST requests when req.body is undefined

jrop:
Addresses #128

Sorry it took me so long to create this.

### DIFF
--- a/lib/passport-saml/strategy.js
+++ b/lib/passport-saml/strategy.js
@@ -70,6 +70,9 @@ Strategy.prototype.authenticate = function (req, options) {
     }
   }
 
+  if (req.method == 'POST' && typeof req.body.SAMLResponse == 'undefined')
+    throw new Error('`req.body` undefined for POST request (are you using body parsing middleware?)');
+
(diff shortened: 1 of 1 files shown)


pdspicer:
I think the check needs to be modified a little bit to potentially avoid "cannot read property 'SAMLResponse' of undefined" errors. In addition I think the check as currently written will break the HTTP-POST binding on SAMLRequests

markstos:
Also, please add automated test coverage for this. Further, the error message doesn't exactly match the behavior. The error says: "`req.body` undefined for POST request", but the check is not for this, but for `req.body.SAMLResponse`. It's possible `req.body.SAMLResponse` could be `undefined`, while `req.body` is still defined.

cjbarth:
@jrop We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

jrop:
@cjbarth I've since moved on from the job where I needed this, and will not get around to modifying this PR to address the comments.  😢

PR #166, opened 2016-08-19
Support detached encrypted key

xdmnl:
Updgrade xml-encryption to support detached EncryptedKey (cf: https://www.w3.org/TR/xmlenc-core/#sec-Extensions-to-KeyInfo, 2nd bullet point).
Okta uses this format for its encrypted SAML responses.


pdspicer:
This looks good to me, thanks for the PR and for testing!

PR #171, opened 2016-09-14
Fixes #170: Clarify that the certificate are looking for is:

### LINKED ISSUE #170: clarify signing vs encryption cert in README
From the README docs:

>  you can provide the Identity Provider's certificate using the cert

It would be helpful to clarify whether it's the IDP's _signing_ certificate or _encrypting_ certificate in the case where there are both. Based on the paragraph above this sentence in the same document, I'm guessing it's the signing certificate.

markstos:
1. public, not private
2. a signing certificate and not an encryption certificate

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Authentication requests sent by Passport-SAML can be signed using RSA-SHA1. To s
     privateCert: fs.readFileSync('./cert.pem', 'utf-8')
 ```
 
-It is a good idea to validate the incoming SAML Responses. For this, you can provide the Identity Provider's certificate using the `cert` confguration key:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,7 +117,7 @@ Authentication requests sent by Passport-SAML can be signed using RSA-SHA1. To s
     privateCert: fs.readFileSync('./cert.pem', 'utf-8')
 ```
 
-It is a good idea to validate the incoming SAML Responses. For this, you can provide the Identity Provider's certificate using the `cert` confguration key:
+It is a good idea to validate the incoming SAML Responses. For this, you can provide the Identity Provider's pulbic signing certificate using the `cert` confguration key:
```
andrsnn: I believe it should be `public`.

PR #179, opened 2016-12-15
set default ForceAuth: false

codeNgamer:
For IDP's who dont assume ForceAuthn: false, set explicit attribute to false


pdspicer:
SAML spec indicates this should default to false if not provided, BUT this would probably be a good option to include (as opposed to a fixed value). Would you mind amending your fix to include `forceAuthn` as an option that can be passed in via the constructor (and passed through to SAML)?

cjbarth:
@codeNgamer Do you have any interest in finishing this up?

codeNgamer:
Hey sorry, I've been swamped and haven't used this in a while but I think I can get this done this weekend. Sound good?

cjbarth:
Sounds good to me. Also, I think the value might already be allowed to pass through, so that might be worth checking before writing code. Thanks for giving time to this even tough you haven't used this in a while.

cjbarth:
@codeNgamer We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

PR #186, opened 2017-01-25
Added to populate eppn, sn and givenName from response while testing…

ppandeypragya:
… with Shibboleth testshib.org IdP


pdspicer:
While I think it's a nice feature that passport-saml parses email from the urn:oid, the verify function provides an option to do any additional parsing from identifiers to names of your choosing, thus I would not be in favor of supporting additional mappings here, it can be left up to the implementer.

ppandeypragya:
Thanks Paul, I'm closing this PR.

PR #187, opened 2017-01-27
Use issuer._ if it exists

idris:
For whatever reason, `issuer` was coming through as:
```
{
  _: 'http://theissuer',
  '$': { Format: 'urn:oasis:names:tc:SAML:2.0:nameid-format:entity'
}
```

This PR fixes it, so it will use the value in `_`, similar to what is done elsewhere in `saml.js`

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -662,7 +662,7 @@ SAML.prototype.processValidlySignedAssertion = function(xml, inResponseTo, callb
 
     var issuer = assertion.Issuer;
     if (issuer) {
-      profile.issuer = issuer[0];
+      profile.issuer = issuer[0]._ || issuer[0];
     }
 
     var authnStatement = assertion.AuthnStatement;
@@ -859,7 +859,7 @@ function processValidlySignedPostRequest(self, doc, callback) {
       }
       var issuer = request.Issuer;
       if (issuer) {
-        profile.issuer = issuer[0];
+        profile.issuer = issuer[0]._ || issuer[0];
       } else {
         return callback(new Error('Missing SAML issuer'));
       }


idris:
@bergie are you still maintaining this repo?

bergie:
@ploer is the current maintainer, but see also #162

pdspicer:
@idris I'm picking up a maintainer role, one ask to add on to this PR: can you make `issuerFormat` available on the profile as well (same as is done for `nameIDFormat`) if issuer is present?

cjbarth:
@idris If there is still interest in this, can you please clean this up. @markstos What refinements are you looking for, especially now that we've updated the profile object a little?

markstos:
For this, I think we should return the spec. I can't find any mention of sending an Issuer in a `_` field, so did this case come up because bad data needed to be parsed from a third party, or because this module did a bad job parsing good data? Before patching this, I'd like have an example of a *valid* SAML document which triggers this case.

jnardone:
I think this project has to decide if they are going to follow the spec or work to integrate with major vendors.  If Okta is doing something weird, yeah that sucks, but they have a ton of marketshare and not supporting them is a disservice to anyone using this library.

markstos:
@jnardone No one in this thread has mentioned Okta or any other vendor by name but you.

jnardone:
The linked PR mentions it specifically: https://github.com/bergie/passport-saml/issues/283

markstos:
@jnardone I've attempted to loop in an Okta rep weigh in, via https://github.com/bergie/passport-saml/issues/283#ref-pullrequest-203537788

andkrist:
> 
> 
> I think this project has to decide if they are going to follow the spec or work to integrate with major vendors. If Okta is doing something weird, yeah that sucks, but they have a ton of marketshare and not supporting them is a disservice to anyone using this library.

I'm having this issue with OAM , so it doesn't seem to be purely related to Okta.

markstos:
@andkrist Okta is trying to work with us to resolve the issue. Please see https://github.com/bergie/passport-saml/issues/283#issuecomment-452416160 and contribute example SAML payloads to contribute.   [Which OAM](https://en.wikipedia.org/wiki/OAM) are you referring to?

andkrist:
Oracle Access Manager. I'm doing some debugging, but right now I have to catch the bus. Will provide example in #288.

cjbarth:
@idris #361 has been merged, which should address this issue. Is your concern still valid now that #361 has landed?

markstos:
I'm closing this PR, assuming #361 supercedes it, but leave a comment if you think there's another improvement not covered by that.

PR #203, opened 2017-04-21
Support multiple certs

austburn:


austburn:
Addresses https://github.com/bergie/passport-saml/issues/189

PR #215, opened 2017-05-30
Add support for ProviderName attribute

alvinward:
Section in SAML spec:

https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf

ProviderName [Optional]
Specifies the human-readable name of the requester for use by the presenter's user agent or the
identity provider.


alvinward:
Creating a new pull requests which also fixes the jshint error.

Issue #214, opened 2017-05-22
No support for <EncryptedKey> linked by reference in SAML XML assertion


Sothos:
The case where an encryption key not defined in the first child element of EncryptedData is not handled. See 
[https://www.w3.org/TR/2002/REC-xmlenc-core-20021210/Overview.html#sec-eg-EncryptedKey](https://www.w3.org/TR/2002/REC-xmlenc-core-20021210/Overview.html#sec-eg-EncryptedKey).

The problem starts in saml.js in function `SAML.prototype.validatePostResponse`. The line 
`var encryptedDatas = xpath( encryptedAssertions[0], "./*[local-name()='EncryptedData']");` is wrong, as is it possible to have the key in a second child element of `<EncryptedAssertion>`.

Attached is an example XML.
![saml](https://cloud.githubusercontent.com/assets/5248621/26294336/25359a86-3e80-11e7-8423-e4bc8a54d89a.jpg)

PR #204, opened 2017-04-21
Support multiple certificates in light of IdP cert rotation. Addresse…

austburn:
…s #189

PR #218, opened 2017-05-31
Support multiple and dynamic signing certificates

richjharris:
In our organisation we roll the signing certificate for our ADFS servers every year.  During the roll over period the server provides two certificates which may be valid, so when a SAML response is provided it is valid if it is signed with either of those signatures.

This change has two methods that can be used when `passport-saml` is used by a long running service that will persist over the certificate roll over period.  The first change is to allow the `cert` configuration key to be an array, this allows for the old and new certificates to be checked during the roll over period.  The second change is to allow the `cert` configuration key to be a function that returns the valid certificate or certificates, this allows the service to poll the ADFS server for what the valid certificates are at this moment and update the signing certificates used by `passport-saml` as they change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,18 @@ If you have a certificate in the binary DER encoding, you can convert it to the 
      openssl x509 -inform der -in my_certificate.cer -out my_certificate.pem
 ````
 
+If the Identity Provider has multiple signing certificates that are valid (such as during the rolling from an old key to a new key and responses signed with either key are valid) then the `cert` configuration key can be and array:
+
+```javascript
+    cert: [ 'MIICizCCAfQCCQCY8tKaMc0BMjANBgkqh ... W==', 'MIIEOTCCAyGgAwIBAgIJAKZgJdKdCdL6M ... g=' ]
+```
+
+The `cert` configuration key can also be a function that returns a certificate or array of certificates.  This allows the Identity Provider to be polled for valid certificates and the new certificate can be used if it is changed:
+
+```javascript
+    cert: function() { return polledCertificates; }
+```
+
 ## Usage with Active Directory Federation Services
 
 Here is a configuration that has been proven to work with ADFS:
--- a/README.md
+++ b/README.md
@@ -145,10 +145,10 @@ If the Identity Provider has multiple signing certificates that are valid (such 
     cert: [ 'MIICizCCAfQCCQCY8tKaMc0BMjANBgkqh ... W==', 'MIIEOTCCAyGgAwIBAgIJAKZgJdKdCdL6M ... g=' ]
 ```
 
-The `cert` configuration key can also be a function that returns a certificate or array of certificates.  This allows the Identity Provider to be polled for valid certificates and the new certificate can be used if it is changed:
+The `cert` configuration key can also be a function that receives a callback as argument calls back a possible error and a  certificate or array of certificates.  This allows the Identity Provider to be polled for valid certificates and the new certificate can be used if it is changed:
 
 ```javascript
-    cert: function() { return polledCertificates; }
+    cert: function(callback) { callback(null,polledCertificates); }
 ```
 
 ## Usage with Active Directory Federation Services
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,9 +6,6 @@ node_js:
   - "4.0"
   - "stable"
 
-before_install:
-  - npm install -g npm
-
```
markstos: Please remove this commit. I'm not sure which older version of Node.js it was trying to support, but we probably don't support it anymore. Recently support for Node.js verisons < 4 was dropped

markstos:
Hi, I'm a new co-maintainer trying to help out a little. 

Please reference the portions of the SAML spec that relate to the changes you are making. If your proposed changes are spec-compliant, I'm inclined to accept a version of this, since supporting multiple certs during a transition seems useful.

Also take a look at the latest version of `tests/test.js`. I refactored to create `TEST_CERT` at the top to store a single copy of a cert that was commonly being re-used, so it doesn't need to be copy/pasted throughout the test file, making it harder to read.

At some point soon, we should also think about breaking up the growing test file into multiple files.

markstos:
It would also be helpful if you could explain how this relates to the other multiple certs PR that was pre-existing: #204.   This if branch goes forward, it also needs to updated to resolve conflicts.

richjharris:
@markstos I wasn't aware of #204 until you pointed it out.  It is doing part of what is being done in this PR.  The differences between #218 and #204 are:
- #204 supports 2 certs rather than an arbitrary number of certs (which is fine for my use case)
- #204 uses a new parameter `alt_cert` for the second cert because it "didn't want to introduce type fluidity" and have the `cert` parameter overridden to accept either a string or an array
- #218 allows for certs to be dynamically updated which #214 doesn't support, #214 requires the service to be reinitialised with the new certs

In terms of referencing these changes in the SAML spec I found this:

>  A relying party SHOULD allow for the use of **any** of the included keys. 

when referring to signing/encryption keys in the SAML metadata https://www.oasis-open.org/committees/download.php/56785/sstc-saml-metadata-errata-2.0-wd-05.pdf `2.4.1.1 [E68]`.

I am happy to base this change off #204 or go as is, if you can let me know what your preference is.

.travis.yml
richjharris: Yep, it was breaking versions < 4 on travis

markstos:
@richjharris your approach seems more flexible, although the author of #204, @austburn is welcome to weigh in. 

One thing I noticed in this PR is when using `cert` as a function, it's synchronous, but the suggested use is to poll a server, which would be an `async` operation. So using a callback pattern seems more appropriate there. Or have I missed something about how it's used as a sync function to check for certs?

austburn:
It's been a while since I've been active on this change -- initially I threw together #204 to address a looming cert rollover. #218 seems to provide more flexibility and I like the ability to poll for cert changes :+1: . I'm fine if this change represents the availability to support cert rollover as it's supported my many other SAML libs.

richjharris:
@markstos the way I use the function version of the `cert` option is I have a scheduled task that polls the current valid certs and stores them and in the function which is passed as the `cert` option I return the certificates which were fetched in the poll.  This allows the function to be synchronous.

I've had a look at where `validateSignature` is called and it wouldn't require any breaking changes to make it async rather than sync if you would prefer it to work that way.

richjharris:
@markstos I've rebased this PR from master but I haven't made any changes on making the function async pending your feedback

markstos:
@richjharris Let's make the `cert` function async to support that usage model.It's easy to make your `sync` use-case run `async`, but it's unwieldy if you want to make an async call with a sync API.

richjharris:
@markstos I've pushed a commit to make the `cert` function async.

cadesalaberry:
This update looks great ! I would love to see it merged.

kallelat:
Indeed something I was just looking for. Hope it gets merged soon!

markstos:
@richjharris Assuming my doc-update to reflect the async change looks good, I think this is ready to release. https://github.com/bergie/passport-saml/pull/218/commits/707b31532650650706e1a2d42f0dee4be02b039b

@cadesalaberry, @kallelat thanks for the peer-feedback. It's helpful to know others are looking for the same feature.

richjharris:
@markstos 👍  looks good.

markstos:
Released and published

PR #231, opened 2017-08-22
handle bad privateCert

chriswininger:
trap exceptions thrown by signRequest which can occurr when an invalid
 privateCert is passed to the saml object. Instead of throwing an
 exception pass the error along via the supplied callback

added spec to check that the callback his hit with the error condtion when
  calling requestToUrl on the saml object

Address issue: https://github.com/bergie/passport-saml/issues/230


markstos:
I ran into this issue today. Thanks for the patch.

PR #235, opened 2017-09-18
Use crypto.randomBytes for ID generation

autopulated:
Math.random is not cryptographically secure, and IDs generated with it could potentially be predicted. Use crypto.randomBytes instead.

I kept the length of the generated ID the same, but it might also be worth increasing it if that's OK everywhere this function is used, as 20 hex chars is only an 80 bit ID, for which collisions could conceivably occur with a realistic number of stored IDs.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -94,12 +94,7 @@ SAML.prototype.getCallbackUrl = function (req) {
 };
 
 SAML.prototype.generateUniqueID = function () {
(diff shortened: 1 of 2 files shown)


markstos:
Thanks for the PR. Could you add a test that confirms that the new strings returned are the same length as the old strings?

autopulated:
added test 👍

markstos:
Thanks, released. Changelog is here: https://github.com/bergie/passport-saml/releases

PR #248, opened 2017-11-08
ES6 node-js aware refactoring

jancajthaml:
ES6 node-js aware refactoring, line-of-sight condition restructuralization.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,4 @@
+{
+  "esversion": 6,
+  "asi" : true
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "4.0"
+  - "6.0"
   - "stable"
 
 script:
--- a/lib/passport-saml/index.js
+++ b/lib/passport-saml/index.js
@@ -1,2 +1,2 @@
-exports.Strategy = require('./strategy');
-exports.SAML = require('./saml').SAML;
+exports.Strategy = require('./strategy')
+exports.SAML = require('./saml').SAML
--- a/lib/passport-saml/inmemory-cache-provider.js
+++ b/lib/passport-saml/inmemory-cache-provider.js
@@ -1,106 +1,92 @@
 /**
- * Simple in memory cache provider.  To be used to store state of requests that needs
- * to be validated/checked when a response is received.
+ * Simple in memory cache provider.  To be used to store state of requests that
+ * needs to be validated/checked when a response is received.
  *
- * This is the default implementation of a cache provider used by Passport-SAML.  For
- * multiple server instances/load balanced scenarios (I.e. the SAML request could have
- * been generated from a different server/process handling the SAML response) this
- * implementation will NOT be sufficient.
+ * This is the default implementation of a cache provider used by Passport-SAML.
+ * For multiple server instances/load balanced scenarios (I.e. the SAML request
+ * could have been generated from a different server/process handling the SAML
+ * response) this implementation will NOT be sufficient.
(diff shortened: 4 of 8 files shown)


markstos:
I will not drop support for Node.js 4 now. 

Node.js is still supported as an LTS release until April, 2018: https://github.com/nodejs/Release

Refactoring the code to require a newer Node.js version can re-visited after that, but this issue is being closed for now.

sibelius:
@markstos can we do this now?

maybe we can use babel to transpile, so we can use latest language features

markstos:
@sibelius Yes, we can drop support for Node.js 4.x now. Node 6 is now the oldest LTS release and should be supported until April, 2019. I'm curious about your interest in the project, though. You haven't otherwise been involved with the `passport-saml` project-- why spend the time to refactor it if you aren't otherwise working on it?  Elsewhere in the issue tracker you'll find that someone had an interest in splitting the project into a "saml library" (separate package) with the passport-specific parts separated.  I mention it as another project that might be happening soon that could touch a lot of the code base. 

I lean slightly against using a transpilation step, but am curious to see what exact language features you might have in might that would benefit from adding that complexity to the project. 

CC: @cjbarth

sibelius:
splitting this lib in a core `saml library` would be very helpfull

I'm trying to do `SAML` redirect url on React (frontend), and then send the SAML response to backend  to be validated

so I don't need to use the passport code only the saml core one

sibelius:
you could avoid using `q` and use `async await` that would make the code much easily to read

cjbarth:
For reference, the issue related to splitting this into a Passport-part and a library-part is in #295, where @xylude mentioned the concept and that thread references a fork that purports to do this (https://github.com/Asana/node-saml-lib) which seems to have some contributors active even somewhat recently. I don't currently have a need for a separate library, but perhaps with these interested ones, such work can be accomplished.

My biggest concern is that we now have good support with @markstos shepherding passport-saml. If some other project hosts the library separately, I worry it will get stale like the aforementioned project has. I'm not sure what the opinion of others is of breaking this project into a library that would still be installed as `passport-saml`, but could then be `require`d as just a library if desired.

markstos:
@cjbarth, 

If the project is split, both parts *could* have the same maintainers, or possibly different ones. That has not been decided. 

There is some possibility that my company will switch from passport-saml to some  third party solution in the future, like Shibboleth, OneLogin or Okta.  So don't plan around the permanence of my maintainership. 

I think perhaps if the project was split, it might attract even more users, since the userbase wouldn't be limited to passport-users as much. Growing the user base also grows the potential pool of maintainers, so it could be win for the project. 

I *am* concerned about the security of the project and would want to have confidence that any other maintainers were being careful to keep the project spec-compliant and secure.

sibelius:
what can use babel preset env and have different targets for different node versions

we can have a target for node 8 that won't transpile async await, and another one that transpile

check this https://babeljs.io/docs/en/babel-preset-env

it could be something like

'passport-saml/modern', so you opt in to use async await natively

markstos:
@sibelius That does sound like  a good solution. It seems we could compile for Node 6, and give Node 8+ users the option to use native code.

cjbarth:
I might also propose that we wait for April 2019 to land this and just drop Node 6 support at that time since that is when LTS support ends. That removes the complexity associated with Babel, and 6 months isn't that long, especially given how long this PR has been active anyway.

markstos:
@cjbarth That approach is OK with me. Given the usually slow pace of changes in this project, someone could go ahead and start on the work if they are OK with the need to possibly rebase and refactor it later to account for changes.

PR #253, opened 2017-12-27
Audience validation

benjamine:
This adds the optional check for SAML AudienceRestriction (see README).
- if `Audience` doesn't match the expected `options.audience` the SAML response is rejected
- if `options.audience` is specified, and SAML response doesn't have an `AudienceRestriction` it is rejected


markstos:
Published today in 0.32.1.

benjamine:
amazing, thanks @markstos !!

PR #256, opened 2018-01-16
Adding a method to profile object to get the actual SAML response

UdayVishwakarma:
I used this library to sign in users to my application using ADFS. I wanted to implement a feature where in users would use AWS `assume-role` feature to get temporary credentials to query for AWS resources. When I need to make api call to AWS for `assume-role` ([assume-role-documentation](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role-with-saml.html)) I need to pass the full SAML response given by ADFS. `passport-saml`s `profile.getAssertionXml()` would return me only assertion xml. where in `profile.getSamlResponse` will be returning the whole `SAML` response

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -309,8 +309,7 @@ SAML.prototype.requestToUrl = function(
       return callback(new Error("Unknown operation: " + operation));
     }
 
-    var samlMessage = request
-      ? {
+    var samlMessage = request? {
           SAMLRequest: base64
         }
       : {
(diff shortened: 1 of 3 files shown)


markstos:
The idea of this PR is fine, but it is impossible to review because it mixes the actual logic change with whitespace changes, jshint changes, replaces how quotes work and adding a `package-lock` file which seems to add unrelated dependencies like `aws4`. 

Please force-push a cleaned up PR which includes *only* the minimal logic changes to add the requested features. Adding test coverage and documentation will also increase the likelihood of acceptance. If you feel compelled to otherwise clean up the syntax and style of the project, please open a new PR for that.

markstos:
Closing in favor of #301  Thanks to @cjbarth for surfacing this as a related PR.

PR #257, opened 2018-01-30
Multiple callback urls

acandylevey:
Multiple callback urls can be now be defined and will be generated within the metadata.xml
Only the default callback is used when sending AuthnRequests. If no default is set, the first option is used as the default.

Can either be set as an array of objects or an array of strings. Existing implementations can still just assign a singular string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ passport.use(new SamlStrategy(
 #### Config parameter details:
 
  * **Core**
-  * `callbackUrl`: full callbackUrl (overrides path/protocol if supplied)
+  * `callbackUrl`: full callbackUrl (overrides path/protocol if supplied); multiple can be supplied as either an array of strings, or an array of objects `{ callbackUrl: 'string', default: true, metadataDefault: true }`. If the default can't be determined the first one will be used. 
   * `path`: path to callback; will be combined with protocol and server host information to construct callback url if `callbackUrl` is not specified (default: `/saml/consume`)
   * `protocol`: protocol for callback; will be combined with path and server host information to construct callback url if `callbackUrl` is not specified (default: `http://`)
   * `host`: host for callback; will be combined with path and protocol to construct callback url if `callbackUrl` is not specified (default: `localhost`)
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -93,7 +93,7 @@ SAML.prototype.getCallbackUrl = function (req) {
   if (this.options.callbackUrl && this.options.callbackUrl.constructor === Array) {
     // Multiple callback urls have been defined use the default one, or if not found - The first one.
     if(this.options.callbackUrl.filter(function (urlOptions) { return urlOptions.default; }).length === 1) {
-      return this.options.callbackUrl.filter(function (urlOptions) { return urlOptions.default; }).callbackUrl;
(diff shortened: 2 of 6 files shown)


markstos:
What is the use-case for defining an array of callbacks vs a single one?

Also, this PR would be more likely to be accepted if it was cleaned up. You could use `git rebase -i` to re-order and combine commits so you're submitting "cleanup", "fix-up" and "woops" commits.

anewtonlevey:
The standard allows for having multiple callbacks. It may be worth
enhancing this so that when sending a request you can choose which indexed
callback you want to send.

I'm specifically using this for when one organisation has multiple tenants
and domains and is connecting via our SSO platform into mutliple systems as
well environments, by having multiple callbacks instead of relaystate this
allows ADFS to be configured for automatic NTLM login to the system and
environment with an IDP intiated sign-on that the organisation wants.

markstos:
> The standard allows for having multiple callbacks. 

Could you link to where this mentioned in the relevant standard?

> I'm specifically using this for when one organisation has multiple tenants
and domains and is connecting via our SSO platform into mutliple systems as
well environments, by having multiple callbacks instead of relaystate this
allows ADFS to be configured for automatic NTLM login to the system and
environment with an IDP intiated sign-on that the organisation wants.

That sounds interesting, but declaring multiple callback seems like a partial solution.

 * How do you make sure the additional callbacks are routed to the right saml configuration to handle it?
 * How do you handle multiple outgoing URLS to redirect to multiple SAML targets?
 * Does your solution publish multiple callback URLs in the same metadata file?

Some docs or an example would help clarify your approach.

I took a different approach to support multiple SAML targets. I modified `passport-saml` to allow renaming the strategy:

https://github.com/RideAmigosCorp/passport-saml/commit/594eed35cc0765172d17487b3611eddffaadd2ea

From there, I was able to create multiple, complete, unique instances of `passport-saml`, each with its own configuration including outgoing redirect URLS, incoming callback URLs and unique metadata.

But if multiple callback URLs are standardized, I think that's also a reasonable approach to support, I'll just admit I'm not clear yet how they are used.

anewtonlevey:
What I'm specifically doing is a bit clunky - But I have the case where I generate a single metadata file for each organisation that we connect into as each of those organisations will have a unique list of multiple callbacks those callback URLs define which resource provider they eventually get passed to. 

So I'm mapping N organisations to N resource providers (We connect to one side via SAML, then authenticate into other resource providers with OAuth2). The resource providers may not be available for every organisation.

Generally, you wouldn't have multiple callbacks of the same binding in the metadata file - You would normally have a POST/REDIRECT/ARTIFACT bindings depending on what you supported. The indexes are provided to allow the IDP to choose (When using IDP initiated sign-on) which callback URL to use if a default isn't specified, although some IDP's do allow the administrator to select which specific callback to use.

Support for allowing multiple callback URLS is expressed in the standard on page 20 - section 2.4.4 under: AssertionConsumerService "One or more".

https://www.oasis-open.org/committees/download.php/35391/sstc-saml-metadata-errata-2.0-wd-04-diff.pdf

cjbarth:
@anewtonlevey We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

PR #265, opened 2018-02-27
Add: Documentation on ADFS errors.

cadesalaberry:


markstos:
Let me know when the PR is no longer a work-in-progress and I'll consider it for merging.

cadesalaberry:
@markstos This documentation is more complete. Tell me what you think.

Cheers,

docs/adfs/README.md
```diff
@@ -111,6 +111,16 @@ module.exports = passport;
 ```
 # Troubleshooting
 
+## Errors you might encounter
+
+|Error   | Explanation  | Mitigation  |
+|--------|---|---|
+| `SAML assertion not yet valid`  | This means that there is a clock skew between your server time and your saml provider time. There might be a better way by fiddling with the ClockSkew You can disable clock skew validation with a value of `-1`.  | `{ acceptedClockSkewMs: -1 }`  |
```
markstos: I think it's helpful to have a chart of possible problems, with explanations and mitigations, so thanks for starting that. Some comments:

 * SAML should be also be written as "SAML", not "saml".
 * Checking the clock skew is an important security feature. Without the check, there is increased vulnerability to reply attacks. So simply advising users to disable the security check is poor advise. First, you should try to fix the clocks on your server and the providers to match UTC closely if possible. If that's not possible, accepting the minimum clock skew between the servers is the nex best option. Disabling the clock skew should be a last resort.

docs/adfs/README.md
```diff
@@ -111,6 +111,16 @@ module.exports = passport;
...
+| `SAML provider returned Requester error: InvalidNameIDPolicy`  | Setting the   |   |
```
markstos: Is there a reference in official SAML docs about the use the Name ID policies that can be linked to? Again, just disabling a policy that's causing a problem doesn't seem like the best solution.

markstos:
I'm rejecting this PR because the advice is generally bad. 

* Checking the timing is an valuable security feature, yet you recommend disabling it completely 
* There a number of potentially valid Name ID formats [as seen here](https://social.technet.microsoft.com/wiki/contents/articles/4038.ad-fs-2-0-how-to-request-a-specific-name-id-format-from-a-claims-provider-cp-during-saml-2-0-single-sign-on-sso.aspx), but rather than recommend finding the correct stricter format to use, you recommend setting the "don't care" format. 

This advice may solve problems in short term, but there may be better options including make sure the clocks are aligned and the correct NameID format is known.

cadesalaberry:
Hi @markstos,

Thanks for your review. I understand your comment and do agree with you. TBH, this table served as a reference to debug the connection with my providers.

The settings can be quite complex, and starting with a 'simple' config allowed me to enable the restrictions one by one.

The message is not clear in my PR, I definitely do not advise disabling those in production.

Would you like me to rephrase? Do you think this table adds value to the documentation?

Cheers,

markstos:
@cadesalaberry I think I'm going to pass on this for now.  

Personally, I keep another document of private/internal SAML documentation that details how we troubleshoot specific errors that come up in our own context.

Our wiki could be a better place to work on collaborative documentation:
https://github.com/bergie/passport-saml/wiki

Issue #252, opened 2017-12-14
RequestedAuthnContext: Comparison type


tjbeers:
Comparison type for RequestedAuthnContext is currently fixed on 'exact'. It would be great if it was configurable as:
* better
* exact
* maximum
* minimum

markstos:
If there's interest in this, first link to the part of the SAML spec where this is behavior is mentioned.  

Second, provide a complete PR to implement it, including code, test and docs.

There are no maintainers here waiting to implement feature requests for others.

tjbeers:
The attribute is specified in the SAML 2.0 Oasis docs (pdf) at: 3.3.2.2.1 Element <RequestedAuthnContext>

I might put some effort into creating a PR, but not before issue #181 is accepted and merged.

tjbeers:
Here's the link to the docs I was referring to:
https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf

markstos:
@tjbeers #298 has been merged with multiple authentication context support and is planned to be replaced soon. Are you or @cjbarth interested in adding support for the comparison options supported by the SAML spec?

cjbarth:
At this time, I'm not.

tjbeers:
Same. But I'm not saying never

PR #270, opened 2018-03-19
Fix for issue #252 RequestedAuthnContext: Comparison type.

ehsankhfr:
Fix for issue [#252 RequestedAuthnContext: Comparison type](https://github.com/node-saml/passport-saml/issue/252).

The "comparisonMethod" is added as a new parameter for config. Details:

 - Bug fix
 - Doc update
 - Tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,15 @@
 node_modules/
 .tern-port
+
+# =========================
+# Package manger lock file
+# =========================
+
+package-lock.json
+yarn.lock
+
+# =========================
+# Webstorm
+# =========================
+
+.idea
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ passport.use(new SamlStrategy(
   * `acceptedClockSkewMs`: Time in milliseconds of skew that is acceptable between client and server when checking `OnBefore` and `NotOnOrAfter` assertion condition validity timestamps.  Setting to `-1` will disable checking these conditions entirely.  Default is `0`.
   * `attributeConsumingServiceIndex`: optional `AttributeConsumingServiceIndex` attribute to add to AuthnRequest to instruct the IDP which attribute set to attach to the response ([link](http://blog.aniljohn.com/2014/01/data-minimization-front-channel-saml-attribute-requests.html))
   * `disableRequestedAuthnContext`: if truthy, do not request a specific authentication context. This is [known to help when authenticating against Active Directory](https://github.com/bergie/passport-saml/issues/226) (AD FS) servers.
+  * `comparisonMethod`: optional comparison method used to evaluate the requested context classes or statements, one of 'exact' (default), 'minimum', 'maximum', or 'better'.
   * `authnContext`: if truthy, name identifier format to request auth context (default: `urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport`)
   * `forceAuthn`: if set to true, the initial SAML request from the service provider specifies that the IdP should force re-authentication of the user, even if they possess a valid session.
   * `providerName`: optional human-readable name of the requester for use by the presenter's user agent or the identity provider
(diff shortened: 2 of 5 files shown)


lib/passport-saml/config/schema.js
```diff
@@ -1,8 +1,8 @@
-const SCHEME = {
+var SCHEME = {
```
markstos: It would be helpful if the commit message explained why it was necessary to replace `const` with `var` here.

ehsankhfr: With node ver4.0, Travis fails to pass. In reality:
`'const' Declarations in non-strict mode are not supported yet on Node v4.`

For the sake of consistency (like the other files), I prevented the use of strict mode.

ehsankhfr: Thanks for the great points! 👍 

I'd rather create a migration branch, to have things more transparent. Can we continue with the fix as it is, and consider that migration as a separate task?

.gitignore
```diff
@@ -1,2 +1,15 @@
 node_modules/
 .tern-port
+
+# =========================
+# Package manger lock file
+# =========================
+
+package-lock.json
+yarn.lock
```
markstos: yarn.lock and package-lock.json *do* belong in Git-- at least one or the other. Distributing them insures that users install exactly the same set of dependencies that the maintainers tested with. This adds consistency which avoids strange bugs. So please remove this from the PR.

If you don't want them to show up as changed locally for you, you can use: https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

Please use amend and rebase to remove the change rather adding a new commit to the branch.

README.md
```diff
@@ -64,6 +64,7 @@ passport.use(new SamlStrategy(
   * `acceptedClockSkewMs`: Time in milliseconds of skew that is acceptable between client and server when checking `OnBefore` and `NotOnOrAfter` assertion condition validity timestamps.  Setting to `-1` will disable checking these conditions entirely.  Default is `0`.
   * `attributeConsumingServiceIndex`: optional `AttributeConsumingServiceIndex` attribute to add to AuthnRequest to instruct the IDP which attribute set to attach to the response ([link](http://blog.aniljohn.com/2014/01/data-minimization-front-channel-saml-attribute-requests.html))
   * `disableRequestedAuthnContext`: if truthy, do not request a specific authentication context. This is [known to help when authenticating against Active Directory](https://github.com/bergie/passport-saml/issues/226) (AD FS) servers.
+  * `comparisonMethod`: optional comparison method used to evaluate the requested context classes or statements, one of 'exact' (default), 'minimum', 'maximum', or 'better'.
```
markstos: Rename this to `requestedAuthContextComparison` to clarify what's being compared.

markstos:
I've reviewed this and it appears to be a spec-compliant feature addition that appears correct, documented and tested. Once the comments I've left above are addressed, it's ready to merge.

.gitignore
markstos: Refinement: `yarn.lock` has been removed from the project, but `package-lock.json` is intended to be commited.

lib/passport-saml/saml.js
```diff
@@ -8,9 +8,11 @@ var querystring = require('querystring');
 var xmlbuilder = require('xmlbuilder');
 var xmlenc = require('xml-encryption');
 var xpath = xmlCrypto.xpath;
-var InMemoryCacheProvider = require('./inmemory-cache-provider.js').CacheProvider;
 var Q = require('q');
```
cjbarth, line 11: This should be rebased on `master` before code review continues.

cjbarth (changes requested):
This need to be rebased off the head of `master` before a proper review can be done.

ehsankhfr:
Thanks for your review, seems the current migration has already done the job of this PR (https://github.com/node-saml/passport-saml/blob/86781395c9c38dd75cbc98aaa562de8a95c225b5/src/passport-saml/saml.ts#L184)

As a result, I close this PR.

PR #274, opened 2018-04-10
Add xml signature to HTTP POST AuthnRequest

tmoiron:
As per SAML RFP, the signature in a HTTP-POST have to be included inside the xml.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -208,7 +208,36 @@ SAML.prototype.generateAuthorizeRequest = function (req, isPassive, callback) {
       request['samlp:AuthnRequest']['@ProviderName'] = self.options.providerName;
     }
 
-    callback(null, xmlbuilder.create(request).end());
+    var generatedRequest = xmlbuilder.create(request).end();
+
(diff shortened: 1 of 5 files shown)


markstos:
When you say "per SAML RFP", please provide a specific link or section reference to the section being implemented or corrected. 

Thanks for the contribution. 

Note there are recent changes to support multiple AuthnRequests which may impact this work.

markstos:
@tmoiron Since it appears you are increasing our spec-compliance, your contribution is welcome here

joedjc:
@tmoiron @markstos Many thanks for doing this it will be really useful. Something we have noticed is that at the moment it is generating the signature at the end of the SAML request, but to comply with SAML schema the Signature needs to be the second element in the request. The services we've tested on are strict on this and reject the request. Otherwise this seems to be working well for logging in. From what I can tell it is not possible to use HTTP-POST binding for sign out at the moment?

markstos:
I believe #206 is a competing pull request that fulfills this goal was merged. If there's some interest from this PR which remains un-addressed by the other PR, leave a comment. In the meantime, I'm closing this.

PR #275, opened 2018-04-10
Adds signing key in the metadata service provider generation.

tmoiron:
As the SP signing key is mandatory to verify signature, it has to be included in the generated metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,11 +173,14 @@ app.get('/login',
 );
 ```
 
-### generateServiceProviderMetadata( decryptionCert )
+### generateServiceProviderMetadata( decryptionCert, signingCert )
+
 
 As a convenience, the strategy object exposes a `generateServiceProviderMetadata` method which will generate a service provider metadata document suitable for supplying to an identity provider.  This method will only work on strategies which are configured with a `callbackUrl` (since the relative path for the callback is not sufficient information to generate a complete metadata document).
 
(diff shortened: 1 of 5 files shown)


markstos:
Seems like a reasonable update. Please admend your commit to document the new parameter you added for metadata generation in the README.

https://github.com/bergie/passport-saml#generateserviceprovidermetadata-decryptioncert-

tmoiron:
Documentation modification performed in README file.

markstos:
Is this backwards compatible with existing configurations? I'm looking particularly at this comment:

> The `signingCert` argument should be a public certificate matching the `privateCert` and is required if the strategy is configured with a `privateCert`.

So, say  I had a privateCert configured, but I had  been calling this method before without the new signingCert argument. Those calls would start to fail after this change, right? 

That could be OK because logically the signing cert is really needed in the metadata in that case, we would just need to do a major version bump in that case. 

Also, could your squash your commits into one and rebase on to master? ( `git rebase -i origin/master` if you have the main repo set as "origin" remote).

markstos:
CC: @cjbarth in case you want to comment.

tmoiron:
> So, say I had a privateCert configured, but I had been calling this method before without the new signingCert argument. Those calls would start to fail after this change, right?

Sure. It fails while calling the generateServiceProviderMetadata function.

>Also, could your squash your commits into one and rebase on to master? ( git rebase -i origin/master if you have the main repo set as "origin" remote).

I'll do that, probably today or tomorrow.

PR #277, opened 2018-04-12
Support redirect for Logout flows

stavros-wb:
• Builds on [#16 Inflating redirect response](https://github.com/bergie/passport-saml/pull/16)
  • Reverts [638ce6e4 accepting redirect as well as post; redirect not verified(638ce6e4e257d99bf1f5b59c2b76f0890744ded9)
  • Implements [
Support HTTP-Redirect binding for IdP to SP SLO](https://github.com/bergie/passport-saml/issues/191)


berzniz:
I would love to see this merged - did you try it with SLO of OneLogin

stavros-wb:
I had a bug when retrieving the `Issuer` from the `LogoutRequest`. I've tested with SimpleSamlPhp and Onelogin. Although I can't find how to setup OneLogin to include a signature for the `LogoutRequest` to make sure the signature validation works as well.

markstos:
To add review, you could link to the part of the SAML spec that discuss SP SLO and related topics.

I think this will be a nice feature to add, but want to be sure about our spec compliance. 

Thanks.

markstos:
Also could you say more about why the mentioned commit was reverted?

markstos:
@stavros-wb Thanks! I'll look for the revised PR.

markstos:
@stavros-wb I'm thinking of doing a "1.0" release within the next 10 days and could include this revised PR if it's ready.

stavros-wb:
thanks @markstos , rebased and squashed on latest master

markstos:
@stavros-wb This looks good to me. Are they are related updates to the README that need to be made for this?

@cjbarth Any other concerns here before this gets merged?

markstos:
@stavros-wb Also, can you confirm if this is considered a breaking change?

cjbarth:
I don't have any concerns about this, but I also don't have a lot of experience with this. I do feel that a README update is warranted though.

stavros-wb:
@markstos 
> @stavros-wb Also, can you confirm if this is considered a breaking change?

Nope. this is not a breaking change. It supports redirects on Logouts, while right now passport-saml supports only `POST`

@cjbarth 
> I don't have any concerns about this, but I also don't have a lot of experience with this. I do feel that a README update is warranted though.

I don't think a README entry is required. I'm not opposed to it, though. SLO is already supported if the protocol binding is `POST` by providing the callback url to the IDP. changing it to redirect would only require changing the setting on the IDP, no further configuration on SP. 
What do you have in mind for the docs?

markstos:
I think it would be useful to mention in the Logout section of the docs that we support SLO and explicitly mention that we support both POST and Redirect bindings.

stavros-wb:
@markstos I've added a small list regarding SLO in the README file

PR #278, opened 2018-04-26
Get session duration from SAML assertion

stavros-wb:


cjbarth:
There are actually a whole bunch of things like this that could be added to the generated profile object. For example there are `Assertion.Conditions` and `Assertion.AuthnStatement` that have very useful properties. I'd be willing to add them, but I'm not sure how we want to handle the idea that there are multiple `Assertion.Conditions` and `Assertion.AuthnStatement`.

Having said that, anyone can get any property from the XML they want by doing parsing `profile.getAssertionXml()`. What do you think an appropriate threshold for data is @markstos ?

cjbarth:
@markstos The problem with option 2 is that it wouldn't be much different from calling `profile.getAssertionXml()` because that returns a JavaScript object anyway. What is more, we already have option 1.

I feel like there is some research that could be done to figure out what the common, or expected, SAML return values are and then build a more complete stable `profile` object around that. @stavros-wb, Do you know of a source for that kind of information? I'd love to see a sample or proposal of what an ideal `profile` object would look like.

I found the schema, which seems a little hard to grok: http://docs.oasis-open.org/security/saml/v2.0/saml-schema-assertion-2.0.xsd
There is also Wikipedia: https://en.wikipedia.org/wiki/SAML_2.0

Sadly the schema is filled with `minOccurs`, but not `maxOccurs`, so we can't really reliably determine how many of many valuable properties there are.

markstos:
I work with dozens of SAML IdPs and typically all that is passed around is the first name, last name and email. An "employee Id" is the next most popular field in my anecdotal experience, but far less popular. There lack of consistency in what the "employee ID" field is named, but the SAML standard term is "NameID". Commonly the unique ID identifier is the email address.  Someone who works at Okta, OneLogin, or similar could provide valuable prospective here.

stavros-wb:
I'm in favour of option 2, but add everything in a property `rawJSON` in the profile. I've tried using `profile.getAssertionXml()` but it returns the XML in a String 
```
> profile.getAssertionXml().constructor
[Function: String]
```
 @markstos I agree we should document this. I would use it, had I known it existed

cjbarth:
@markstos @stavros-wb I've created a PR (#301) that follows the spirit of this discussion. Please have a look.

markstos:
@stavros-wb Does #301 address your interest here? It's a more a general solution to the same issue.

The one difference I see is that this PR adds `profile.sessionDuration` to the profile object, while in #301, I think the same data would have to accessed as:

  profile.getAssertion().SessionNotOnOrAfter

stavros-wb:
yes #301 will do the trick. thank you both @cjbarth @markstos . closing

PR #289, opened 2018-07-13
added namespace md: for metadata elements

urbanenomad:
Thanks for creating this lib.  I am using it for an automated service which reads the metadata output.  Unfortunately, your lib does not provide the namespace prefix for the root metadata namespace.  So I updated the metadata function to insert the "md:" prefix.  I already accept other metadata xml files such as from modules like simplesaml and they provide the md: namespace prefix which allows my service to parse the metadata file to auto provision trust with our IDP.  I hope this is useful for others and I hope it doesn't break any one elses code.


urbanenomad:
sorry I missed place one of the elements

PR #290, opened 2018-07-13
Add metadata namespace prefix "md:" when generating metadata.xml

urbanenomad:
Apolgize for my earlier mistake.  I think this code has the proper position of the metadata elements.  This again is to help my automation process for configure trust with our IDP service.  This would place the namespace of the metadata prefix in the front of the metadata generated xml.


markstos:
First, Please provide a reference to the SAML spec that shows where this change is spec-compliant behavior.

Second, please rebase and force-push to this branch so there is a single, clean commit.

Thanks.

cjbarth:
@urbanenomad We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

Issue #181, opened 2016-12-29
How to specify multiple authnContext while configuring saml stratergy?


Dhiviya-Bharathi:
In my project, we have 2 authentication methed - Password Protected Transport and Integrated Windows Authentication. 

How should we specify both of them in the config?

We tried as comma seperated value as below, but didn't work.

authnContext: 'urn:federation:authentication:windows,urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport'

PR #298, opened 2018-09-11
support multiple authnContext

cjbarth:
This addresses issue #181 and supersedes PR #183. Please see those other two entries for context.


markstos:
Looks good. Thanks. Merging.

QA Log
=======
  *  I confirmed that SAML 2 support multiple authentication contexts. [ref](https://msdn.microsoft.com/en-us/library/ee895365.aspx)

cjbarth:
As soon as this lands I'll update the TypeScript typings for this. Is there any interest in maintaining the typings for this project here?

markstos:
@cjbarth Only if you maintain them. My team plans to adopt the TypeScript in the future, but hasn't currently. The project is changing at a slow rate now, so maintenance to the TypeScript bindings should be infrequent.

cjbarth:
In that case, I'll leave them over at DefinitelyTyped for now.

Issue #284, opened 2018-05-31
Identify SP-initiated vs Idp-initiated SAMLResponse


suchitagarwal:
When I receive I samlResponse, I would like to find out whether the response is solicited `InResponseTo` a SAMLRequest sent by my server, or was it unsolicited and initiated at the IdP. 

I was thinking of looking at the presence of `InResponseTo` in the samlResponse, but it looks like that is not added by default. 

I could call `getAssertionXml`, parse the xml and check for presence of `InResponseTo`. However, given that passport-saml already checks this, would it be possible to just add this id to the response sent to the caller? 

If not, are there any other clean ways of identifying whether a response was initiated at the IdP or at the SP?

rupertbg:
Hey @suchitagarwal I'm trying to figure out the same thing. Currently I'm using `validateInResponseTo: true` but it doesn't not seem to validate if the attribute does not exist at all.. Is this potentially a bug?

rupertbg:
I'm no expert but is this saying "Only do the validation if the attribute doesn't exist"? https://github.com/bergie/passport-saml/blob/c2ce79d51d93b68e34911e74bb06cf915d8a754b/lib/passport-saml/saml.js#L556-L564

markstos:
> would it be possible to just add this id to the response sent to the caller?

Providing a code example of what you have in mind would be helpful.

markstos:
This being addressed by #300, so closing this issue.

PR #300, opened 2018-09-13
Handle case of missing InResponseTo when validation is on

cjbarth:
This supersedes PR #286 and addresses issue #284. Please references those items for context.


cjbarth:
I've also included a stub JS file that will help future users to sign sample XML used for creating tests.

PR #302, opened 2018-09-25
Handle case of missing InResponseTo when validation is on

markstos:


markstos:
This supercedes #300 by @cjbarth, by amending his commit to correct the spelling of "signing" in a new file name.

PR #304, opened 2018-09-25
feat(multi): fix multi saml strategy export

### LINKED ISSUE #303: can't use MultiSamlStrategy
```
Error: Cannot find module 'passport-saml/MultiSamlStrategy'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
```

sibelius:
multi saml strategy import won't worked well

fix #303

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ passport.use(new SamlStrategy(
 You can pass a `getSamlOptions` parameter to `MultiSamlStrategy` which will be called before the SAML flows. Passport-SAML will pass in the request object so you can decide which configuation is appropriate.
 
 ```javascript
-var MultiSamlStrategy = require('passport-saml/MultiSamlStrategy');
+var MultiSamlStrategy = require('passport-saml').MultiSamlStrategy;
 [...]
 
 passport.use(new MultiSamlStrategy(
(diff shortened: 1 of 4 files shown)


markstos:
The previous design was intentional: It meant the multi-Saml code was only loaded if it was needed, rather than always-loaded as it is in the new design.  The typo was not intentional.

sibelius:
I think this does not make much sense on server side, as the bundle side does not change anything

markstos:
@sibelius It is a very minor memory optimization to not load the file if it's not needed. It won't make a practical difference in most cases. In my case, I'm running about 100 Node processes on two twin servers which may each end up loading passport-saml. Any memory savings are multiplied by 200x in this case. A few small changes may help put off the need to upgrade to larger hardware just a little bit further.  

So I'm resolving this with the original intent by correcting the typo in the README instead of changing the code. 

https://github.com/bergie/passport-saml/commit/25ac94e37d17b371eaa90f344a5f0d17a162d5ad

That said, I appreciate your efforts to put together a PR to improve the project.

PR #308, opened 2018-10-03
Fixes #296

### LINKED ISSUE #296: Unexpected deletion of Strategy config.name property during creation
I recently updated from `0.32.1` to `0.35.0`. This caused by application to fail with the following error message
`Authentication strategies must have a name`

After investigation, I found the following code was added in `0.33.0` https://github.com/bergie/passport-saml/blob/v0.33.0/lib/passport-saml/strategy.js#L17:L23

```
  // Customizing the name can be useful to support multiple SAML configurations at the same time.
  // Unlike other options, this one gets deleted instead of passed along.
  if  (options.name) {
    this.name  = options.name;
    delete options.name;
  }
  else {
    this.name = 'saml';
}
```
I was already using a name property in the config so that I can setup multiple SAML strategies, sample code shown below

```
    let samlStrategy = new SAMLStrategy(conf, (profile, done) => {
      // some code here
    });

    passport.use(conf.name, samlStrategy);
```

The above code now fails because for the change made in `0.33.0` and I had to solved this issue by making a clone of the `conf` object and passing the clone when creating the strategy.

Question: Shouldn't the module take care of creating the clone if it is going to delete properties of parameters that is been passed to it?

josnidhin:
Also updates the supported node engine min version to 6. This is required to use the new API used by this fix

### DIFF
--- a/lib/passport-saml/strategy.js
+++ b/lib/passport-saml/strategy.js
@@ -6,6 +6,11 @@ function Strategy (options, verify) {
   if (typeof options == 'function') {
     verify = options;
     options = {};
+  } else {
+    // make a local copy of the options object
+    // refer https://github.com/bergie/passport-saml/issues/296
+    // for details
+    options = Object.assign({}, options);
   }
 
   if (!verify) {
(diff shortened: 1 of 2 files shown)


josnidhin:
Forgot to update travis config

PR #311, opened 2018-10-05
feat(logout): handle null and undefined on nameQualifier

sibelius:
microsoft outlook returns

`nameQualifier`
`spNameQualifier`

as `null` instead of `undefined`

so this PR address this to handle both null and undefined


markstos:
I manually tested the logic before and after several values: undefined, null, '', 'string', and 0.  

I found that the only change in the logic was to treat both "undefined" and "null" the same way, as intended, I'm merging this.

Some automated tests here could be helpful to prevent a related regression in the future.

PR #314, opened 2018-10-15
Validate issuer on logout requests/responses if configured

stavros-wb:
I've added documentation on `samlIssuer` and also made its validation optional (just as `InResponseTo`). This should actually be a part of https://github.com/bergie/passport-saml/pull/277

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,8 @@ type Profile = {
   * `validateInResponseTo`: if truthy, then InResponseTo will be validated from incoming SAML responses
   * `requestIdExpirationPeriodMs`: Defines the expiration time when a Request ID generated for a SAML request will not be valid if seen in a SAML response in the `InResponseTo` field.  Default is 8 hours.
   * `cacheProvider`: Defines the implementation for a cache provider used to store request Ids generated in SAML requests as part of `InResponseTo` validation.  Default is a built-in in-memory cache provider.  For details see the 'Cache Provider' section.
+ * **Issuer Validation**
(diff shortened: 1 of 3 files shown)


stavros-wb:
@markstos can you take a look?

README.md
```diff
@@ -125,6 +125,8 @@ type Profile = {
   * `validateInResponseTo`: if truthy, then InResponseTo will be validated from incoming SAML responses
   * `requestIdExpirationPeriodMs`: Defines the expiration time when a Request ID generated for a SAML request will not be valid if seen in a SAML response in the `InResponseTo` field.  Default is 8 hours.
   * `cacheProvider`: Defines the implementation for a cache provider used to store request Ids generated in SAML requests as part of `InResponseTo` validation.  Default is a built-in in-memory cache provider.  For details see the 'Cache Provider' section.
+ * **Issuer Validation**
+  * `samlIssuer`: if provided, then the IdP issuer will be validated for incoming Logout Requests/Responses
```
markstos: I think it could be helpful to add an example at the end of this line that shows an example expected value for `samlIssuer`.  The name here is also confusing, since we have both `issuer` and `samlIssuer`. 

As I understand `issuer` is *our name*-- the name of the Service Provider, while `samlssuer` is the expected name of the IdP that we validate against. Is that correct?  Perhaps 'responseIssuer` would be clearer?

Including as "saml" as a prefix does not help much at all, because the entire module has to do with SAML.

stavros-wb: I see, that's a valid point, but `responseIssuer` isn't really better, since this is validated in requests as well.

What do you think for `idpIssuer`?

markstos: Perfect. Thanks.

stavros-wb: done

PR #317, opened 2018-10-30
Manually applying changes from fork to workaround inclusive namespaces

couilllard45682:
Source: https://github.com/mindspacepdx/passport-saml/pull/1


couilllard45682:
Please delete. Doesn't work well. 
https://github.com/auth0/xml-crypto/commit/25a3d1c00fd0751530f7a06f7219c6eb6860412d does work better for that case.

PR #341, opened 2019-01-24
Remove InResponseTo value if response validation fails

Archinowsk:
InResponseTo value is used to verify that the response is for a valid request. If InResponseTo doesn't match a request the response is invalid. This is a cheap verification before more expensive cryptographical validation.

If cryptographical validation fails the response is most likely invalid and the corresponding InResponseTo value should be removed so no further calls can be made using it.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -721,7 +721,14 @@ SAML.prototype.validatePostResponse = function (container, callback) {
   })
   .fail(function(err) {
     debug('validatePostResponse resulted in an error: %s', err);
-    callback(err);
+    if (self.options.validateInResponseTo) {
+      Q.ninvoke(self.cacheProvider, 'remove', inResponseTo)
+      .then(function() {
+        callback(err);
+      });
+    } else {
+      callback(err);
+    }
   })
   .done();
 };
--- a/test/tests.js
+++ b/test/tests.js
@@ -535,13 +535,13 @@ describe( 'passport-saml /', function() {
                 var query = response.headers.location.match( /^[^\?]*\?(.*)$/ )[1];
                 encodedSamlRequest = querystring.parse( query ).SAMLRequest;
               }
-  
+
               var buffer = Buffer.from(encodedSamlRequest, 'base64');
               if (check.config.skipRequestCompression)
                 helper(null, buffer);
               else
                 zlib.inflateRaw( buffer, helper );
-  
+
               function helper(err, samlRequest) {
                 try {
                   should.not.exist( err );
@@ -772,7 +772,7 @@ describe( 'passport-saml /', function() {
           try {
             var id = doc['samlp:LogoutRequest']['$']["ID"];
             var issueInstant = doc['samlp:LogoutRequest']['$']["IssueInstant"];
-  
+
             id.should.be.an.instanceOf(String);
             issueInstant.should.be.an.instanceOf(String);
             cacheSaveSpy.called.should.eql(true);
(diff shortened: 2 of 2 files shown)


markstos:
@josecolella Would you mind peer-reviewing this?

lib/passport-saml/saml.js
```diff
@@ -704,7 +704,15 @@ SAML.prototype.validatePostResponse = function (container, callback) {
...
   .fail(function(err) {
     debug('validatePostResponse resulted in an error: %s', err);
-    callback(err);
+    if (self.options.validateInResponseTo) {
+      Q.ninvoke(self.cacheProvider, 'remove', inResponseTo)
+      .then(function() {
+        callback(err);
+      });
+    }
+    else {
```
cjbarth: nit: The rest of the code uses `} else {` instead of putting that on two lines.

markstos:
@Archinowsk or @cjbarth Could you link the part of the SAML spec that deals with InResponseTo validation, for reference?

Archinowsk:
Squashed commits and fixed formatting as @cjbarth suggested.

I tried running tests against master and the same tests fail.


Is this ok for specs?

http://docs.oasis-open.org/security/saml/v2.0/saml-sec-consider-2.0-os.pdf

"6.1.3 Message Insertion

Threat: A fabricated request or response is inserted into the message stream. A false response such as a spurious “yes” reply to an authorization decision query or the return of false attribute information in response to an attribute query may result in inappropriate receiver action. 

Countermeasures: The ability to insert a request is not a threat at the SOAP binding level. The threat of inserting a false response can be a denial of service attack, for example returning SOAP Faults for responses, but this attack would become quickly obvious. The more subtle attack of returning fabricated responses is addressed in the SAML protocol, appropriate since according to the SOAP Binding definition each SOAP response must contain a single SAML protocol response unless it contains a fault. The SAML Protocol addresses this with two mechanisms, correlation of responses to requests using the required **InResponseTo** attribute, making an attack harder since requests must be intercepted to generate responses, and through the support origin authentication, either via signed SAML responses or through a secured transport connection such as SSL/TLS."

markstos:
I want to make sure I understand this by explaining this in my own words:

`InResponseTo`  should only be used for one request/response cycle.

This patch is addressing a case where a response fails cryptographically. When this happens, the updated code will remove the related `InResponseTo` from the internal cache.

The benefit is that future attempts to respond using the same InReponseTo value will fail faster because `InResponseTo` is checked before the cryptographic check.

Are we on the same page about that?

However, it looks like the last release accidentally reverted an improvement to `InResponseTo` validation. See my note at: https://github.com/bergie/passport-saml/commit/7674e18e#r32746468

I think the other issue needs to be addressed as part of merging and releasing this. 

And yes, there are some unrelated tests failing on master that need to be addressed. A separate PR is welcome to address those. It looks like there was a bad merge recently. 

CC: @josecolella

Archinowsk:
Yes, you described the logic correctly.

cjbarth:
@Archinowsk I find that there is a clean merge of `master` into this branch and all the tests pass. If @markstos would restart the Travis CI build again, I think the merge would succeed.

Otherwise, I can't write to your branch, so I merged `bergie:master` into `Archinowsk:remove-inresponseto-if-response-validation-fails` and pushed that to `cjbarth:remove-inresponseto-if-response-validation-fails`, so you can feel free to add my source as an origin and then `reset --hard` to `cjbarth:remove-inresponseto-if-response-validation-fails` and push that up to `Archinowsk:remove-inresponseto-if-response-validation-fails` which should kick off this CI job again and get everything working again.

Also, I see that your clone of this project has lots of little changes. Is there any interest on your part in getting those pushed upstream for us all to benefit from? If you need help, please let me know.

markstos:
@cjbarth Thanks for helping move things along!

Archinowsk:
Thanks, @cjbarth! Pushed changes and the build is passing now. Ended up cleaning some whitespaces in the process.

> Also, I see that your clone of this project has lots of little changes. Is there any interest on your part in getting those pushed upstream for us all to benefit from? If you need help, please let me know.

I'm working with a project that has its own fork: https://github.com/vrk-kpa/suomifi-passport-saml. We have been adding suitable fixes to the upstream but there are lots of stuff that would not make sense in a standard implementation :).

test/tests.js
```diff
@@ -968,16 +968,44 @@ describe( 'passport-saml /', function() {
...
+        var samlObj = new SAML( samlConfig );
+
+        // Mock the SAML request being passed through Passport-SAML
+        samlObj.cacheProvider.save(requestId, new Date().toISOString(), function(){});
+
+        samlObj.validatePostResponse( container, function( err, profile, logout ) {
+          should.exist( err );
+          err.message.should.match("Invalid signature");
+          samlObj.validatePostResponse( container, function( err, profile, logout ) {
+            should.exist( err );
```
cjbarth: This `done()` is in a callback, so it needs to be in a `try...catch` block, otherwise a failure here will be reported as a timeout when the tests runs.

Archinowsk: Good catch, fixed!

There was also one Buffer() remaining so I updated it to Buffer.from().

PR #350, opened 2019-02-08
Support InResponseTo validations in MultiSaml

stavros-wb:
Either use cache provided by user, or a default memory
cache to store InResponse parameters. This cache is not
yet partitioned per provider, which means a malicious
provider could do replay attacks by using anothers
unconsummed `InResponse` values

https://github.com/bergie/passport-saml/issues/334

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,11 @@ passport.use(new MultiSamlStrategy(
   })
 );
 ```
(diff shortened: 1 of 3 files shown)


stavros-wb:
@markstos PTAL

PR #353, opened 2019-02-14
feat: allow SPSSOdescriptor attributes configuration

sami-sweng:
This PR allows the user to configure three attributes of SPSSODescriptor.

From the [SAML 2 SP metadata specifications](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-metadata-2.0.xsd) we got `AuthnRequestsSigned` and `WantAssertionsSigned`.

`WantAssertionsEncrypted` seems to not be part of the specification and be implementation dependent, but some organizations configure their IdP to require it.


cjbarth:
@sami-sweng We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

PR #354, opened 2019-02-14
Feat/allow consumer service index configuration

sami-sweng:
This PR allows user to configure the index of `AssertionConsumerService` in the generated metadata.

The [specifications](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-metadata-2.0.xsd) only specify that it should be an `unsignedShort`, but some IdP implementations requires it to start at `0`.


cjbarth:
@sami-sweng We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

PR #355, opened 2019-02-14
Convert cert to pem if not already.

VilleMiekkoja:
Docs require that cert has no header and footer lines. But verify function in code requires them

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -778,6 +778,7 @@ function processValidlySignedSamlLogout(self, doc, callback) {
 }
 
 SAML.prototype.hasValidSignatureForRedirect = function (samlMessageType, container) {
+  var self = this;
   var signature = container.Signature;
   if (signature && this.options.cert) {
     var urlString = samlMessageType + '=' + encodeURIComponent(container[samlMessageType]);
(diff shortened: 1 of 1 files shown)


lib/passport-saml/saml.js
```diff
@@ -778,6 +778,7 @@ function processValidlySignedSamlLogout(self, doc, callback) {
 }
 
 SAML.prototype.hasValidSignatureForRedirect = function (samlMessageType, container) {
+  var self = this;
```
markstos: `self = this` is not necessary if arrow functions are used, which are supported on all Node versions that we support.

lib/passport-saml/saml.js
```diff
@@ -792,7 +793,7 @@ SAML.prototype.hasValidSignatureForRedirect = function (samlMessageType, contain
     return this.certsToCheck()
       .then(function(certs) {
         var hasValidQuerySignature = certs.some(function (cert) {
-          return validateSignatureForRedirect(
+          return self.validateSignatureForRedirect(
             urlString, signature, container.SigAlg, cert
```
markstos: Instead of using `function (cert) {`, you can use an use an arrow function here: `cert => {`. 

Arrow functions do not create a new `this` binding, so you can use `return this.validateSignatureForRedirect` here and it will refer to the outer `this`.

PR #356, opened 2019-02-14
InResponseTo support for logout

VilleMiekkoja:


markstos:
@josecolella Since you are looking at some other aspects of logout, what do you think of this PR?

josecolella:
@markstos The code makes sense. If the tests pass, and delineate the enhancement. I think we can merge this

PR #360, opened 2019-03-05
feat: add RequestedAuthnContext Comparison Type parameter

osan15:


osan15:
The modifications have been made

markstos:
I've merged this, but renamed `comparisonType` to `RACComparison`, as `comparisonType` is too generic. I also documented in the main README.

Issue #283, opened 2018-05-23
Issuer is different in samlResponse for different IdP


suchitagarwal:
Hi,
I am implementing SAML SSO using passport-saml. I want to be able to support all IdPs like GSuite, Okta, Onelogin, Auth0. I had been testing it with GSuite till now and everything works fine. However, when I tested with Okta, I noticed that there is a difference in the samlResponse that we get in the callback.

For GSuite, the samlResponse looks like:

```js
{
  nameID: user@example.com,
  issuer: https://accounts.google.com/…,
  …
}
```

for Okta, it looks like: 

```js
{
  nameID: user@example.com,
  issuer: {
    _: 'https://example.okta.com/oktaid/…',
    other data here…
  },
  …
}
```

Effectively, the issuer is a string for GSuite samlResponse, while its an Object for an Okta samlResponse. Should this be handled by passport-saml itself, is this a known issue?

a-legrand:
I agree, other fields are handled, for example ```profile.nameID = nameID[0]._ || nameID[0];``` but not the Issuer

cjbarth:
@suchitagarwal @a-legrand , #187 seems to address this. There is a note on that PR asking for documentation before it gets landed. Do either of you have some documentation that could shed light on the standards compliance of this pattern?

suchitagarwal:
@cjbarth not off the top of my head, let me look around to see if i can find something

markstos:
@mraible You work at Okta. Could you connect us with a SAML developer there who can shed light on why Okta returns their SAML responses in a different format than G-Suite? We want the `passport-saml` project to be spec-compliant, but also compatible with major providers like Okta. See ticket above for context. Thanks.

jpf:
Hey folks, I'm a co-worker of @mraible at Okta. Do you have some examples of the XML payloads that you're getting from the various IDPs?

Below is an example of what I'm looking for, it's a freshly minted and lightly anonymized SAML Response from one of my Okta orgs. To my eyes, the Issuer looks as it should be, but I'd like to compare this response with similar responses from other SAML IdPs:

```xml
<saml2p:Response xmlns:saml2p="urn:oasis:names:tc:SAML:2.0:protocol"
                 Destination="https://example.okta.com/auth/saml20/jfranusic"
                 ID="id10703141429762751406940984"
                 IssueInstant="2019-01-08T19:04:37.100Z"
                 Version="2.0"
[...]

PR #361, opened 2019-03-07
Set explicitChar: true to make XML parsing consistent. Fixes issue #283 and #187

andkrist:
By setting explicitChars to true whenever xml2js is used and always use ._ to access values incosistent checks for [0] and [0]._ are eliminated.

Not broken test that were not broken from before.

Could have set .$ and ._ to .attrs and .chars or something instead but to fix the inconsistensies in the issues this was not necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 .tern-port
+.idea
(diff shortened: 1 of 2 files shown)


markstos:
The `._` syntax is ugly, but as discussed in #283, I think the consistent XML parsing is worth it here. 

Thanks. Merging.

PR #363, opened 2019-03-18
Update README.md

josecolella:
Change attributeName to unknown type to allow for use-case described in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33950. For example, you may extend the Profile object with an attribute `roles` which is an array of string. Having unknown still allows for typing enforcement, and yet is flexible to allow for these use-cases


josecolella:
(cc: @markstos). This patch is needed to create a PR for the passport-saml types. :)

PR #367, opened 2019-04-26
Fix broken tests

cjbarth:
There were two bugs in the code that were causing the tests to fail. Due to the way the tests were written tests failed as timeouts. I've also adjusted the tests so that when they fail they give useful information.

No functional changes were made to the test. The only changes were to correctly wrap async exception throwing blocks in a `try...catch` and then call `done()` properly so that `mocha` could properly report the error instead of having to wait 2 seconds for timeout and then reporting that instead of the actual error that occured.

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -129,14 +129,18 @@ describe( 'passport-saml /', function() {
           };
 
           request(requestOpts, function (err, response, body) {
-            should.not.exist(err);
-            response.statusCode.should.equal(check.expectedStatusCode);
-            if (response.statusCode == 200) {
-              userSerialized.should.be.true;
-              if (check.expectedNameIDStartsWith)
-                profile.nameID.should.startWith(check.expectedNameIDStartsWith);
+            try {
+              should.not.exist(err);
+              response.statusCode.should.equal(check.expectedStatusCode);
+              if (response.statusCode == 200) {
+                userSerialized.should.be.true;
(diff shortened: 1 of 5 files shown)


cjbarth:
As I was updating typings for v1.1 when it is released I found that there were some inconsistencies and metadata was being mishandled. Error handling was also broken.

test/tests.js
```diff
@@ -2042,52 +2223,76 @@ describe( 'passport-saml /', function() {
...
-          should.not.exist(err);
-          success.should.eql(true);
-          done();
+          try {
+            should.not.exist(err);
+            success.should.eql(true);
+            done();
+          } catch (err2) {
+            done(err2);
+          }
```
markstos: This test does not appear to actually be async, so could be converted to a sync test without try/catch.

markstos:
@cjbarth I see that's a lot of work that's been put here.

I'm reluctant to merge this because the try/catch syntax is verbose and ugly. I also understand that's solving a real problem with sync errors being thrown in async test. 

Now that Node 6.x is EOL'ed, we can drop support for it, which means it's safe to start using `async/await` in the code base.  This could help clean up the tests, but to take advantage of it we should also update some functions to return a promise if no callback is provided.

Still, I realize a release is overdue for for the moment I have gratitude that this fixes for the tests for now, so I'm merging it.

PR #375, opened 2019-05-21
NameIDFormat fix

ahavriluk:
https://github.com/bergie/passport-saml/issues/338

When identifierFormat: null the passport-saml generated metadata with a tag `<NameIDFormat/>` which causes an error in AD FS: `The value 'NameIDFormat' must be an absolute URI`.
Setting the identifierFormat: urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified allows importing the metadata file successfully.


markstos:
@sibelius This look OK to me. What do you think?

PR #379, opened 2019-07-01
Validate signatures on original query string

stavros-wb:
Before this commit:
signature validation on GET (Redirect binding) requests was done using
`encodeURIComponent` on the already parsed params

After:
the original query params are used to validate the request's signature

The reason for doing this is that Azure ADFS is using lowercase for url
encoding resulting in something like:
`SAMLRequest=dead%2fbeef`
The old processing logic would try to validate against:
`SAMLRequest=dead%2Fbeef`
and failed

see also:
https://github.com/onelogin/php-saml/issues/243


markstos:
I want to make sure I understand the root cause here. Is it:

   A. "%2f" and "%2F" aren't being decoded the same. *If that's the issue, it's a bug in the decoder, because [URI percent encoding should not be case-sensitive](https://stackoverflow.com/questions/7994287/is-url-percent-encoding-case-sensitive)
  B. Because the URL has changed, the signature changed, so signature validation fails. This seems like a bug in Azure ADFS, but one we may want to workaround because Azure ADFS is popular.

stavros-wb:
> I want to make sure I understand the root cause here. Is it:
> 
> A. "%2f" and "%2F" aren't being decoded the same. _If that's the issue, it's a bug in the decoder, because [URI percent encoding should not be case-sensitive]_(https://stackoverflow.com/questions/7994287/is-url-percent-encoding-case-sensitive)

they're decoded ok, the issue is after the decode we do an encode to validate the signature. this encode can't derive if the original had lowercase or uppercase encodings. to amend I'm using the same parameters

>B. Because the URL has changed, the signature changed, so signature validation fails. This seems like a bug in Azure ADFS, but one we may want to workaround because Azure ADFS is popular.

Azure just happens to forward lowercase encodings, while the encodeURI produces only uppercase ones

lib/passport-saml/saml.js
```diff
@@ -837,7 +837,9 @@ SAML.prototype.hasValidSignatureForRedirect = function (container, originalQuery
 
 function validateSignatureForRedirect (urlString, signature, alg, cert) {
   var supportedAlgs = crypto.getHashes().filter(
-    function(h) { return new RegExp(h).test(alg); }
+    function(h) { return new RegExp(h, 'i').test(alg); }
```
markstos: The "i" has been added here because URL encoding is not case sensitive, is that correct?

stavros-wb: no this is another issue. if azure is configured with `rsa-sha256` (lowercase) signing it would match with `sha256` crypto algorithm. doing a case incentive search in the crypto algorithms resolves this.
```js
crypto.getHashes()
[ 'RSA-MD4',
  'RSA-MD5',
  'RSA-MDC2',
  'RSA-RIPEMD160',
  'RSA-SHA1',
  'RSA-SHA1-2',
  'RSA-SHA224',
  'RSA-SHA256',
  'RSA-SHA384',
  'RSA-SHA512',
  'blake2b512',
  'blake2s256',
  'md4',
  'md4WithRSAEncryption',
  'md5',
  'md5-sha1',
  'md5WithRSAEncryption',
  'mdc2',
  'mdc2WithRSA',
  'ripemd',
  'ripemd160',
  'ripemd160WithRSA',
  'rmd160',
  'sha1',
  'sha1WithRSAEncryption',
  'sha224',
  'sha224WithRSAEncryption',
  'sha256',
  'sha256WithRSAEncryption',
  'sha384',
  'sha384WithRSAEncryption',
  'sha512',
  'sha512WithRSAEncryption',
  'ssl3-md5',
  'ssl3-sha1',
  'whirlpool' ]
```

lib/passport-saml/saml.js
```diff
@@ -837,7 +837,9 @@ SAML.prototype.hasValidSignatureForRedirect = function (container, originalQuery
...
+  ).sort(
+    function(h) { return -h.length; }
```
markstos: Why are we even testing with a RegExp here? Wouldn't it be safer and less error-prone to test with an exact match instead? I can see possibly testing case-insensitive.   

But if the algorithm used was `mdc2`, shouldn't we use the `mdc2` algorithm and not the `mdc2WithRSA` algorithm?

The way I read this code, it tests with a case-insensitive substring match and then chooses the algorithm with the longest name that matches. 

We also throw away the entire array but the last entry after running a regex over over hash and then sorting it. 

It would be clearer and faster to stop looking as soon as find the match we we want. 

We could skip using `supportAlgs` at all. The current code was using a case-sensitive match and it seemed to work fine. If we test for exact matches, then the incoming `alg` can be passed through to `createVerify()` assuming it's an algorithm supported by Node.js

stavros-wb: the current version was also taking the last entry:
```js
function validateSignatureForRedirect (urlString, signature, alg, cert) {
  var supportedAlgs = crypto.getHashes().filter(
    function(h) { return new RegExp(h).test(alg); }
  );
  if (supportedAlgs.length === 0) {
    throw alg + ' is not supported';
  }
  var verifier = crypto.createVerify(supportedAlgs[supportedAlgs.length-1]);
  verifier.update(urlString);
  return verifier.verify(cert, signature, 'base64');
}
```

markstos:
> no this is another issue. if azure is configured with rsa-sha256 (lowercase) signing it would match with sha256 crypto algorithm. doing a case incentive search in the crypto algorithms resolves this.

I see. Is there any reason not to use the more precise approach of doing a case-insensitive lookup?  

Thanks for the Azure fixes. One of the next customers I need to integrate with uses Azure.

stavros-wb:
> I see. Is there any reason not to use the more precise approach of doing a case-insensitive lookup?

a stricter mapping might be the right thing here, but I didn't want to break any existing implementations (even if they might be working by accident)

markstos:
Some more follow-up here: I found that algorithms arrive as URLs, so an exact match approach does not work, even case-insensitive. I committed an approach that was more precise than a basic substring match, though.  An example algorithm that arrives in an SAML response looks like this: `http://www.w3.org/2001/04/xmlenc#sha256`. I used a Regex match to extract the part after the `#` character. *Then* we try an exact-but-case-insensitive match against all the algorithms that Node.js supports. 

Still, we could do strong validation. I looked up what valid algorithms are for SAML and how the names are formatted. SAML is using XML Encryption and XML Signatures, which each have their own W3C specs:

 * https://www.w3.org/TR/xmldsig-core/#sec-Algorithms
 * https://www.w3.org/TR/2002/REC-xmlenc-core-20021210/Overview.html#sha256

So, the algorithms we should *actually* support would be some sub-set of the algorithms defined there. 

**We could make signature validation even stronger by testing incoming algorithms against the list of algorithms supported in the XML Signatures spec**. Since we don't do that now, it's possible that we could accept an algorithm that is weaker than was even intended to be used.

PR #387, opened 2019-09-04
Private Certificate should be called Private Key

RyanCopley:
I could be wrong, but I have never heard of a private certificate before. This confused my team quite a bit and i think it would be useful to correct this. I'm not claiming this is the optimal implementation, but starting this conversation would be great.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -151,7 +151,7 @@ SAML.prototype.signRequest = function (samlMessage) {
     samlMessageToSign.SigAlg = samlMessage.SigAlg;
   }
   signer.update(querystring.stringify(samlMessageToSign));
-  samlMessage.Signature = signer.sign(this.options.privateCert, 'base64');
+  samlMessage.Signature = signer.sign(this.options.privateCert || this.options.privateKey, 'base64');
 };
 
 SAML.prototype.generateAuthorizeRequest = function (req, isPassive, callback) {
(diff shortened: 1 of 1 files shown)


louielang:
I agree with Ryan. I was super confused when I started using passport-saml and I was struggeling with this. At least the docs should be updated so its clear what "privateCert" is really for.

markstos:
@louielang a PR to improve the docs is welcome.

cjbarth:
I was also confused when I started using this project, but I too am concerned about backward compatibility. Perhaps we can consider this a breaking change for v2?

markstos:
@RyanCopley The way you've implemented this, it IS backwards compatible, so it doesn't require a major semver release. Would you be willing to update the patch to be compatible with our current TypeScript code and add some docs and test for it? 

This project is maintained by other users like you. It's unlikely anyone else is going to finilsh this feature for you if you want to see this happen.

RyanCopley:
Oh man, I've already switched jobs and tech stacks entirely now :( It may be hard for me to find time to do this since I won't be given company time, and I don't have any projects that have this problem anymore.

RyanCopley:
I'll try contacting my former employer and see if I can get them to take a look.

alon85:
> I'll try contacting my former employer and see if I can get them to take a look.

Please review this updated PR: https://github.com/node-saml/passport-saml/pull/488

markstos:
Closing in favor of #488

PR #388, opened 2019-09-05
use options.host for callbackUrl

warncke:
When callbackUrl was not specified passport-saml would use the HOST value from the request (req.headers.host) instead of the host specified in options. This change uses the host from options unless it is "localhost" which is the default value for that option. In the case of "localhost" we will continue to use the req.headers.host value as before.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -103,12 +103,8 @@ SAML.prototype.getCallbackUrl = function (req) {
   if (this.options.callbackUrl) {
     return this.options.callbackUrl;
   } else {
-    var host;
-    if (req.headers) {
-      host = req.headers.host;
-    } else {
-      host = this.options.host;
-    }
+    // Use host from options unless it is the default (localhost)
+    var host = this.options.host && this.options.host !== 'localhost' ? this.options.host : req.headers.host;
     return this.getProtocol(req) + host + this.options.path;
   }
 };
--- a/test/samlTests.js
+++ b/test/samlTests.js
@@ -32,6 +32,47 @@ describe('SAML.js', function () {
       };
     });
 
+    describe('getCallbackUrl', function () {
+      describe('when callbackUrl and host/path not specified', function () {
+        it('returns callbackUrl based on request headers', function () {
+          var callbackUrl = saml.getCallbackUrl(req);
+          callbackUrl.should.equal('https://examplesp.com/saml/consume')
+        });
+      });
+
+      describe('when protocol, host, path specified', function () {
+        beforeEach(function () {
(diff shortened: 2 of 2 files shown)


markstos:
It does seem strange and possibly problematic to trust the incoming request to generate an outgoing request. 

@warncke Could you reference the SAML spec and see what it says to about this behavior, if anything?

cjbarth:
@warncke I'd be happy to look at this more closely when I get time. It would certainly help reduce the time I need to find if you could reference the SAML spec about this.

markstos:
I reviewed the spec here. This relates to our generation of the "AssertionConsumerServiceURL" as a service provider. Our use of incoming HTTP data looks safe here because we reference "req.headers.host, which is the "Host:" header which is used to route the request to the right virtualhost to respond. If an attacker tried to fake this header, I believe it would route the request to a different virtualhost entry at the web server and fail to respond. There is perhaps a narrow window for mischief if the app responds to two different virtual hosts and there is some value in changing the host header from one value to another that the app answers to. 

Regarding this specific proposal, the way I read it there is a desire to not put "localhost" in the AssertionConsumerServiceURL values that we generate. Presumably this would would help with some cases of local testing.

But there is another way to accomplish the same thing which is generally useful for development: If you want to pretend that your laptop is a different domain, add an entry for the domain to `/etc/hosts` and map that domain name to 127.0.0.1. Using that method, this change is not needed. 

I'm conservative about making changes to this module. Given that no one else has asked for this feature and there appears to be a workaround, I'm closing the issue.

Now, if there appears to be a _security_ issue with the current behavior that we've missed please point that out.

warncke:
Sorry for the delay in responding on this. There is a bit of a miscommunication here I think about what the intent of this change is.

The only intent here is to address an inconsistency with how configuration works.

The readme defines the host option in the config as "host for callback; will be combined with path and protocol to construct callback url if callbackUrl is not specified (default: localhost)".

As you can see from the code change, the existing code does not use the host value from the configuration, and does not default to localhost, but instead defaults to the value in the host header in the request.

The fact that you can specify what you want the callbackUrl to be using one configuration option and not the other, even though the documentation specifically says that the other can be used, seems like a simple implementation inconsistency to me.

markstos:
>  the existing code does not use the host value from the configuration, and does not default to localhost

The existing code _does_ default to options.host, which in turn defaults to 'localhost' for the case when req.headers is not defined.

Unless there's a problem with the current behavior, it would be preferred to address the inconsistency between docs and code by updating the docs for `host` to clarify that it defaults to to the host header when constructing the callback URL and `req.headers` is present.

PR #389, opened 2019-09-05
add useFriendlyNames config option

warncke:
Adds a new `useFriendlyNames` config option that when set to true will use the `FriendlyName` instead of `Name` for profile properties from the SAML assertion.

Fixes https://github.com/bergie/passport-saml/issues/282

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ type Profile = {
   * `skipRequestCompression`: if set to true, the SAML request from the service provider won't be compressed.
   * `authnRequestBinding`: if set to `HTTP-POST`, will request authentication from IDP via HTTP POST binding, otherwise defaults to HTTP Redirect
   * `disableRequestACSUrl`: if truthy, SAML AuthnRequest from the service provider will not include the optional AssertionConsumerServiceURL. Default is falsy so it is automatically included.
+  * `useFriendlyNames`: if truthy `FriendlyName` from assertion will be used instead of `Name` for profile property name
  * **InResponseTo Validation**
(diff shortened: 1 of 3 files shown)


markstos:
I'm not opposed, but FriendlyName is optional in the spec and the name is required. 

One benefit would be making configurations a bit more readable by referencing the FriendlyName. On the other hand, since FriendlyName is optional, depending on it being there could increase the chance  of problems. 

I'm leaving this open in case someone else wants to comment for or against.

cjbarth:
@warncke I'm not against this either, but even if the option is set, there should be logic to fallback to `Name` in the case that `FriendlyName` doesn't exist.

markstos:
@cjbarth There is logic to fallback to "name" if FriendlyName does not exist: https://github.com/bergie/passport-saml/pull/389/commits/50024ffa0c8f7b89ef9dd5831c25964e2c0a3d28#diff-41fc97a3e28a2b6f8b5ec3bdae6130d5R1034 

But that seems that could lead to confusion there the resulting profile names are a mix of the "Friendly" and "Non-friendly" names. 

What if we add a "friendly" key to the profile returned and make the "friendly" names sub fields of that? It would be up to the end user to default to the non-friendly names if they want to do that.

cjbarth:
Ah yes @markstos , I must have missed that due to the wrap. Sorry. But I do agree that we could have implementer confusion if we mix them in the library. It does make more sense to have multiple keys and let the implementer sort it out.

markstos:
I'm rejecting this because as proposed the result can be a mix of "friendly" and non-friendly names based on the incoming data which can sew confusion in an already complex library.

Friendly names are appealing but because they are optional they can lead to complications.

PR #390, opened 2019-09-05
convert privateCert to PEM for signing

warncke:
allow `privateCert` to be passed in as base64 string without PEM formatting the same as `cert`.


markstos:
I might use this myself. Please amend your commit to update the docs as well and I think I'll merge this. 

The version of `systemd` on Ubuntu 18.04 doesn't like newlines in environment variables, so this will be a welcome update for those who want to store the PEM in a one-line environment variable.

cjbarth:
@warncke I second @markstos comments. Please update the docs and lets get this landed.

warncke:
@markstos @cjbarth sorry for the long delay on this. I have update the readme to try to accurately reflect the change.

PR #396, opened 2019-10-01
Fix jshint errors

walokra:
'const' is available in ES6
Missing semicolon.
'let' is available in ES6


markstos:
Thanks, but this was resolved by dropping support for Node 6.

PR #397, opened 2019-10-01
Fix #392 Switching from jshint to eslint

### LINKED ISSUE #392: Switch from jshint to eslint
Eslint seems powerful and flexible and I'm more familiar with it. It seems we don't even have an .jshint file now, so we are just using  the default rules.

walokra:
Used rules with exceptions what were allowed when using jshint and enforcing using semicolons.
Eslint gives error when linting es6 features as rules are enforcing ecmaVersion: 5.
Eslint 6 requires Node 8 or higher.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,10 @@
+env:
+  node: true
+  es6: false
+extends: 'eslint:recommended'
+parserOptions:
+  ecmaVersion: 5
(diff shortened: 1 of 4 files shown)


markstos:
Thanks. This has been refined, merged and pushed. I dropped support for Node 6, so we could upgrade our linting support to ES6.

PR #400, opened 2019-10-02
Bring-up xml-crypto to 1.4.0

LoneRifle:
This update will pick up the following changes:

yaronn/xml-crypto#171
yaronn/xml-crypto#179
yaronn/xml-crypto#183

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "debug": "^3.1.0",
     "passport-strategy": "*",
     "q": "^1.5.0",
-    "xml-crypto": "^1.1.4",
+    "xml-crypto": "^1.4.0",
     "xml-encryption": "^0.11.0",
     "xml2js": "0.4.x",
     "xmlbuilder": "^11.0.0",


markstos:
@LoneRifle I see you appear to be a commiter for `xml-crypto`. I appreciate the nudge here, but do these changes fix known bugs in `passport-saml`?

cjbarth:
I don't know of any reported bugs to this project that this would fix, however https://github.com/yaronn/xml-crypto/issues/164, which is fixed in this proposed update, covers a scenario that could affect users of this library.

cjbarth:
Side note, it seems the CI build failed because we are specifying testing be done with an out-dated version of Node.

markstos:
@cjbarth That looks like a legit reason to upgrade. I'm bumping this dependency. 

Regarding the CI failures due to old Node.js versions. Are you referring to the .travis.yml file? We bumped the Node.js version is that about a month ago: https://github.com/bergie/passport-saml/blob/master/.travis.yml#L3

cjbarth:
So we did. Consider this closed then.

PR #401, opened 2019-10-02
Follow up on dd1699a - Travis to use Node 8

LoneRifle:


walokra:
This is a duplicate of #399.

PR #402, opened 2019-10-06
Fix #355 missing parts: tests.

walokra:
Add changes for PR #355 "Convert cert to pem if not already" which were requested and were done in suomifi-passport-saml fork.

Note: self = this is needed and tests fail if using arrow function as suggested in the PR review.


markstos:
Thanks.

Issue #279, opened 2018-05-03
Does this lib have any substantial logging functionality?


aliaksander-samuseu:
Hi.

I'm dealing with an custom Passport SAML strategy which relies on `passport-saml` lib. What is puzzles me is that I don't see any way how to find out what is happening under the hood of `passport-saml` when it does its job. If, for example, a SAML response will come in which it won't be able to handle (for any reason), it just silently fails with ambiguous error messages. I have zero idea of what was the cause etc.

When I check sources of `passport-saml` I don't see that much logging instructions anyway. This is very strange as it makes troubleshooting virtually impossible, and imo should be a mandatory feature of any such piece of code.

May be I'm missign something? How is this problem usually tackled?

markstos:
A patch is welcome for an optionally enabled `debug` move to improve logging. The `debug` module may help with this. A "Troubleshooting" or "Debugging" section could also be added to the README.

markstos:
Starting with 0.35.0, released today, you can set `DEBUG='passport-saml'` in the environment to get more debug logging. Right now this only works for the `validatePostResponse` function, but a pull request is welcome to extend debug logging elsewhere. In 0.34.0, also released today, diagnostic messages were improved for the "Invalid Signature" cases, you can be certain which of the six invalid signature cases you triggered.

PR #407, opened 2019-12-05
improve debug output (#279)

giladwolff:
This commit improves the debug output by adding the SAML response ID to
the debug message and by exposing the validation errors from xml-crypto
in case signature validation fails.

Here are examples of the improve output:

passport-saml checkSignature failed for pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa. Validation errors: invalid signature: for uri #pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa calculated digest is 0AbCoTZl3NxNBiPUyucHk/7gay8= but the xml to validate supplies digest DCnPTQYBb1hKspbe6fg1U3q8xn4= +-2017d

passport-saml validatePostResponse for id "_6a377272c8662561acf1056274ef3f81" resulted in an error: Error: SAML provider returned Responder error: InvalidNameIDPolicy +1ms

Testing done:
ran 'DEBUG=passport-saml npm run-script test' successfully,
saw the expected output.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -303,7 +303,7 @@ SAML.prototype.generateLogoutResponse = function (req, logoutRequest) {
 };
 
 SAML.prototype.requestToUrl = function (request, response, operation, additionalParameters, callback) {
-  
+
   const requestToUrlHelper = (err, buffer) => {
     if (err) {
       return callback(err);
@@ -537,7 +537,7 @@ SAML.prototype.validateSignature = function (fullXml, currentNode, certs) {
   if (signatures.length != 1) {
     return false;
   }
-  
+
   const signature = signatures[0];
   return certs.some(certToCheck => {
     return this.validateSignatureForCert(signature, certToCheck, fullXml, currentNode);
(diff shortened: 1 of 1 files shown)


lib/passport-saml/saml.js
```diff
@@ -584,6 +592,11 @@ SAML.prototype.validatePostResponse = function (container, callback) {
     if (!Object.prototype.hasOwnProperty.call(doc, 'documentElement'))
       throw new Error('SAMLResponse is not valid base64-encoded XML');
 
+    var id = xpath(doc, "/*[local-name()='Response']/@ID");
```
markstos: I'm fine with improving debug output, but was concerned that this `xpath()` call might add a new case that could throw. 

I checked that we get the function from `xml-crypto` which gets it from `xpath.js`, which does not document the behavior when a select fails. Turns out I opened a bug to have those docs improved in 2016, but it hasn't happened yet. 

I also confirmed there's no test coverage for that case either:

https://github.com/yaronn/xpath.js/issues/18

As long as you are improving things, could you submit a PR to `xpath.js` as well which tests and documents the behavior if a selection were to fail?

giladwolff: Thanks, I'll do that and report back.

markstos: It looks like that xpath.js project seems pretty dead, but someone has forked it, rewritten it in TypeScript and added more test coverage and docs: https://github.com/backslash47/xpath

Maybe we should convince xml-crypto to use this version instead.  

Oh, the nice new TypeScript fork depends at least Node v11 and our policy is to support the oldest LTS release of Node.js that's supported, which is currently Node.js 10. We we won't drop Node 10 support for over another year.

markstos: Our xpath dependency changed. The correct project to be looking at is: https://github.com/goto100/xpath   It's a fork, but it retains the lack of documentation and test coverage for failing selects. At least this branch is somewhat maintained.

cjbarth:
@giladwolff We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

PR #411, opened 2020-01-10
Correction of docs of host parameter.

kvanrobbroeck:
Changed documentation of `host` parameter to state the default value is the contents of `Host` header, instead of `localhost`


markstos:
I'm not sure this change is quite correct either. In one spot the default value *is* localhost:

https://github.com/bergie/passport-saml/blob/master/lib/passport-saml/saml.js#L36

However, where we actually use the value, the default value varies. _if_ req.headers is present and passed in the default value does come from the Host: header as you suggest.

But otherwise, it defaults to `options.host` which defaults to `localhost`. 

Updated docs should perhaps make the distinction between `localhost` being the default for the `host` option to initialize, and also update the docs to whatever public, documented function calls `getCallbackUrl()` to be clear how the default values for host are set there.

cjbarth:
@kvanrobbroeck We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

PR #412, opened 2020-01-15
Always throw error objects instead of strings

Gekkio:
While JS *technically* allows code to throw anything (including undefined/null/NaN), it's often considered bad practice. It's very annoying to handle/log errors from a library that sometimes throws strings and sometimes objects, and there's also no stack trace in string errors to help with debugging.

I realize this might be a controversial change, and it can break existing error handling code that assumes strings are thrown in some of the cases. But maybe it could be at least considered for 2.0?

More information:

https://eslint.org/docs/rules/no-throw-literal
http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-rejected-with-a-non-error

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -712,7 +712,7 @@ describe( 'passport-saml /', function() {
     it( 'should throw an error if cert property is provided to saml constructor but is empty', function() {
       should(function() {
         new SAML( { cert: null } );
-      }).throw({ message: 'Invalid property: cert must not be empty' });
+      }).throw('Invalid property: cert must not be empty');
     });
 
     it( 'generateUniqueID should generate 20 char IDs', function() {
--- a/test/tests.js
+++ b/test/tests.js
@@ -712,7 +712,7 @@ describe( 'passport-saml /', function() {
     it( 'should throw an error if cert property is provided to saml constructor but is empty', function() {
       should(function() {
         new SAML( { cert: null } );
-      }).throw('Invalid property: cert must not be empty');
+      }).throw({ message: 'Invalid property: cert must not be empty' });
     });
 
     it( 'generateUniqueID should generate 20 char IDs', function() {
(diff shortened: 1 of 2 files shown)


markstos:
I'm aligned this, and agree it would be a "breaking change" for some. 

@Gekkio Could your review what our docs say about the exceptions we throw and make sure they reflect the change as well?

markstos:
I agree with this change. I was going to save it a for a bundled "2.0" release, but clearly that approach didn't work. If you'll address the merge conflicts, I'll try to get it out sooner, which may possibly mean moving away from a bundled 2.0 release to a series of breaking change releases.

Gekkio:
Sorry, I completely forgot about this pull request :sweat_smile: 

I've rebased the changes on top of current master, so the pull request is now up-to-date. I also looked at the docs and couldn't find a detailed description of thrown errors that would need to be changed, so the docs seem ok to me.

test/tests.js
```diff
@@ -712,7 +712,7 @@ describe( 'passport-saml /', function() {
     it( 'should throw an error if cert property is provided to saml constructor but is empty', function() {
       should(function() {
         new SAML( { cert: null } );
-      }).throw('Invalid property: cert must not be empty');
+      }).throw({ message: 'Invalid property: cert must not be empty' });
```
cjbarth, line 715: Based on [this](http://shouldjs.github.io/#assertion-throw), it doesn't seem that we should haven an object here and that the previous syntax should continue to work.

markstos:
@cjbarth Even though we didn't document the error format returned, I still think we should do a major version bump for this change. Would you agree?

cjbarth:
I completely agree. This already has the `2.0` and `breaking-change` tags, so I think we are good to honor that. Once this is totally ready to land we can discuss what else might be good for a `2.0` release.

test/tests.js
Gekkio: True, the extra wrapper was unnecessary! I've removed it

cjbarth (approved):

markstos (approved):

markstos:
We are planning a 2.0 release today, so I'm merging this.

PR #414, opened 2020-01-21
attributeConsumingServiceIndex can be zero

gunzip:
The PR fixes a wrong check on attributeConsumingServiceIndex which can be zero.


cjbarth:
@gunzip This looks reasonable, can you add a test for this?

markstos:
I agree it looks good and a test would also be good. However, because the fix looks obvious I've also going to go ahead and merge this now so it doesn't miss the next release, and still welcome additional test coverage here in a future PR.

PR #420, opened 2020-02-06
Upgrade xml-encryption to 1.0.0

brandon-leapyear:
Currently running into this: https://github.com/auth0/node-xml-encryption/issues/55


markstos:
This patch is not complete because it does not updated yarn.lock.

markstos:
This should have perhaps been treated as a breaking change as upgrading to xml-encryption disables insecure algorithm. (Or it's a bug fix, depending how you look at it!).

PR #423, opened 2020-02-18
Singleline private keys

big-kahuna-burger:
- provides a better implementation of #390


markstos:
Released in 1.3 .3 today.

PR #426, opened 2020-02-26
Fix multi saml strategy race conditions

stavros-wb:
https://github.com/bergie/passport-saml/issues/425

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,9 @@ You can override these defaults by passing a new value through the `getSamlOptio
 
 Using multiple providers supports `validateInResponseTo`, but all the `InResponse` values are stored on the same Cache. This means, if you're using the default `InMemoryCache`, that all providers have access to it and a provider might get its response validated against another's request. [Issue Report](!https://github.com/bergie/passport-saml/issues/334). To amend this you should provide a different cache provider per SAML provider, through the `getSamlOptions` function.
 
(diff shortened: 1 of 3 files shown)


gunzip:
thakn you for this fix ! maybe want to provide a big warning for those still using < 1.3.3 packages from npm ?

stavros-wb:
@gunzip https://github.com/bergie/passport-saml/blob/a2ccde0ccdd802b4b12f2450e55166bf8ab01fab/README.md#configure-strategy-for-multiple-providers

gunzip:
about `< 1.3.3` we should update that when this pr gets merged (to see into which release).
The main problem I see is not that
> a User will authenticate successfully through the wrong IDP
but that any subsequent request which happens concurrently to the login endpoint (not an extreme condition : ) will overwrite the saml client options making the call to authenticate() fail.

stavros-wb:
Yes the simple manifestation of this bug (I agree not an extreme condition) will be an error on a user trying to login, but I don't consider this part of the Readme. There are other open issues as well, that aren't part of the Readme.

The more dangerous manifestation of the bug (which I describe happens under extreme conditions) would be a wrong authentication, a sec vulnerability. 

I think we should document the false positives (more dangerous) and not the false negatives. WDYT?

gunzip:
imho it *should* be considered a vulnerability since any user can DoS the login endpoint preventing any other user to authenticate

stavros-wb:
amended

stavros-wb:
@markstos PTAL

markstos:
Published in 1.3.4.

Issue #245, opened 2017-10-30
attrValueMapper fails to parse complex AttributeValue tags


vpalmisano:
attrValueMapper function parse correctly this AttributeStatement:

`<saml2:Attribute FriendlyName="NAME" Name="urn:oid:x.x.x.x" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri">
  <saml2:AttributeValue>VALUE</saml2:AttributeValue>
</saml2:Attribute>`

but it fails with this kind of AttributeStatement:

`<saml2:Attribute FriendlyName="NAME" Name="urn:oid:x.x.x.x" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri">
<saml2:AttributeValue>
  <saml2:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:persistent" NameQualifier="XXX" SPNameQualifier="XX">VALUE</saml2:NameID>
</saml2:AttributeValue>
</saml2:Attribute>`

huntdm07:
The PR I've submitted #249 should fix this with support for existing usage.

alexstuart:
Hello, it's Alex from the UK federation support team here. Just to expand on what @huntdm07 says in https://github.com/bergie/passport-saml/pull/249. 

The UK federation uses eduPersonTargetedID to uniquely identify subjects. [In SAML 2 eduPersonTargetedID is called `urn:oid:1.3.6.1.4.1.5923.1.1.1.10`](http://software.internet2.edu/eduperson/internet2-mace-dir-eduperson-201602.html#eduPersonTargetedID) and, as far as I know, it's the only attribute from the eduPerson schema which has a value that is a `saml2:NameID`. So I think your code should only be trying to extract parts of the attribute if its Name starts with `urn:oid:1.3.6.1.4.1.5923.1.1.1.` and its `Name == urn:oid:1.3.6.1.4.1.5923.1.1.1.10`.

Secondly, the UK federation advises storing eduPersonTargetedID by concatenating the following elements:
 
1.    the entity name of the identity provider
1.    a single ‘!’ character
1.    the entity name of the service provider
1.    a single ‘!’ character
1.    the opaque string value

As an example, the attribute received as
```
<saml2:Attribute
           FriendlyName="eduPersonTargetedID"
           Name="urn:oid:1.3.6.1.4.1.5923.1.1.1.10"
[...]

PR #249, opened 2017-11-13
 Add support nested NameID in attribute value

huntdm07:
Addressing #245.

This is a common format in IdPs in the UK Federation, including the test IdP (https://www.ukfederation.org.uk/content/Documents/TestIdP), which provide values like the following:

```
<saml2:Attribute
           FriendlyName="eduPersonTargetedID"
           Name="urn:oid:1.3.6.1.4.1.5923.1.1.1.10"
           NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri">
          <saml2:AttributeValue>
                <saml2:NameID
                   Format="urn:oasis:names:tc:SAML:2.0:nameid-format:persistent"
                   NameQualifier="https://test-idp.ukfederation.org.uk/idp/shibboleth"
                   SPNameQualifier="https://test.ukfederation.org.uk/entity">
                  xY1234567890ACdFGGhjmkklljA=
                </saml2:NameID>
          </saml2:AttributeValue>
</saml2:Attribute>
```

PR #246, opened 2017-10-30
Fixing NameID value mapping

vpalmisano:
Fixes #245

PR #427, opened 2020-03-09
fix: attrValueMapper fails to parse complex AttributeValue tags (#245)

dfdeagle47:
# Context

This fixes an issue where the `attrValueMapper` would fail to properly map the value for complex `AttributeValue` tags. This handles the case where the `AttributeValue` contains a nested `NameID` tag.

One such example is the `eduPersonTargetedID` that is used as an identifier in [eduGAIN][1] which can return an Attribute of the form

```xml
<saml2:Attribute FriendlyName="eduPersonTargetedID" Name="urn:oid:1.3.6.1.4.1.5923.1.1.1.10" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri">
  <saml2:AttributeValue>
    <saml2:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:persistent" NameQualifier="https://idp.example-university.fr/idp/shibboleth" SPNameQualifier="https://www.service-provider.com/shibboleth">a6c2c4d4-08b9-4ca7-8ff9-43d83e6e1d35</saml2:NameID>
  </saml2:AttributeValue>
</saml2:Attribute>
```

# Limitations

Note that in reality, the `AttributeValue` tags can be much more complex. [The Assertions and Protocols for the OASIS Security Assertion Markup Language (SAML) V2.0][2] uses the following schema for the `Attribute` tag:

```xml
<element name="Attribute" type="saml:AttributeType"/>
<complexType name="AttributeType">
  <sequence>
    <element ref="saml:AttributeValue" minOccurs="0" maxOccurs="unbounded"/>
  </sequence>
  <attribute name="Name" type="string" use="required"/>
  <attribute name="NameFormat" type="anyURI" use="optional"/>
  <attribute name="FriendlyName" type="string" use="optional"/>
  <anyAttribute namespace="##other" processContents="lax"/>
</complexType>
```

and the following schema for the `AttributeValue`:

```xml
<element name="AttributeValue" type="anyType" nillable="true"/>
```

which means it can take any type.

As pointed out in [this issue][3], it is customary to use `NameQualifier` and the `SPNameQualifier` in addition to the actual value to create a unique identifier for the platform. That is why the `AttributeValue` is mapped to an object containing the attribute of the `NameID` tag as well as the string value for `eduPersonTargetedID` that is stored in the `Value` property.

# Additional information regarding the fix

- The `attrValueMapper` is extracted to the SAML prototype to make it easier to test and it is renamed to `attributeValueMapper` to make it more explicit.
- Tests have been added to check that it correctly handles the use case where the AttributeValue contains a nested NameID tag.
- Previously, the `attrValueMapper` would return `undefined` for this use case. Now, it returns an object where `Value` is the nested string value and the tag attributes are properties of that object.

# Related PRs and issues

This should close the following PRs (although the mapping is different):
- https://github.com/bergie/passport-saml/pull/246
- https://github.com/bergie/passport-saml/pull/249

This should close the following issue:
- https://github.com/bergie/passport-saml/issues/245

[1]: https://wiki.geant.org/display/eduGAIN/Identifier+Attributes
[2]: https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
[3]: https://github.com/bergie/passport-saml/issues/245#issuecomment-344575492

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -1008,10 +1008,6 @@ SAML.prototype.processValidlySignedAssertion = function(xml, samlResponseXml, in
                                                   .map(attr => attr.Attribute)
                                );
 
-      var attrValueMapper = function(value) {
-        return typeof value === 'string' ? value : value._;
-      };
-
       if (attributes) {
         attributes.forEach(attribute => {
          if(!Object.prototype.hasOwnProperty.call(attribute, 'AttributeValue')) {
@@ -1020,9 +1016,9 @@ SAML.prototype.processValidlySignedAssertion = function(xml, samlResponseXml, in
           }
           var value = attribute.AttributeValue;
           if (value.length === 1) {
-            profile[attribute.$.Name] = attrValueMapper(value[0]);
+            profile[attribute.$.Name] = this.attributeValueMapper(value[0]);
           } else {
-            profile[attribute.$.Name] = value.map(attrValueMapper);
+            profile[attribute.$.Name] = value.map(this.attributeValueMapper);
           }
         });
       }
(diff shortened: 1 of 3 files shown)


dfdeagle47:
I still have to fix one issue before merging.

cjbarth:
Since #245 is closed, should this be closed too @dfdeagle47 ?

dfdeagle47:
@cjbarth 
> Since #245 is closed, should this be closed too @dfdeagle47 ?

Looks like it does, thanks for the update.

You might want to look at those PRs as well:

- #246
- #249

because they seem to be fixing the same issue.

PR #435, opened 2020-05-13
Feat/replace q with native promises

gugu:
Improved existing PR and removed Q-shim completely (promisify used instead)


gugu:
Based on #409

markstos:
This is a welcome update, but conflicts need to be addressed now.

PR #436, opened 2020-05-13
Es6

gugu:
Used lebab to convert the code to ES6

It already includes changes from #435


markstos:
@gugu  This now has a lot of conflicts. What do you recommend as next steps?

gugu:
I'd rather close it and send incremental PRs

PR #440, opened 2020-06-17
Added option to set singleLogoutServiceBinding

simonsmadsen:


markstos:
@simonsmadsen Please add test coverage and address the merge conflicts.

README.md
```diff
@@ -148,6 +148,7 @@ type Profile = {
   * `logoutUrl`: base address to call with logout requests (default: `entryPoint`)
   * `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
   * `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
+  * `singleLogoutServiceBinding`: if set to `HTTP-Redirect`, sets the @Binding attribute in the `SingleLogoutService` element to `urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect`. Default is `urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST`.
```
cjbarth, line 151: It isn't clear with this setting if it should be set to just the suffix or to the entire value. I think this could be cleared up if the value had to be set in its entirety. Thus, one would set it to `urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect`. not `HTTP-Redirect`. The code to use this value wouldn't do string concatenation.

lib/passport-saml/saml.js
```diff
@@ -1272,7 +1272,7 @@ SAML.prototype.generateServiceProviderMetadata = function( decryptionCert, signi
 
   if (this.options.logoutCallbackUrl) {
     metadata.EntityDescriptor.SPSSODescriptor.SingleLogoutService = {
-      '@Binding': 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
+      '@Binding': 'urn:oasis:names:tc:SAML:2.0:bindings:'+ this.options.singleLogoutServiceBinding || 'HTTP-POST',
```
cjbarth, line 1275: It seems we should avoid string concatenation here and just require the entire value to be set. That makes the documentation clearer and makes the configuration value in whatever config file much more explicit, which is good for a security system.

cjbarth (changes requested):

lib/passport-saml/saml.js
markstos: Agreed.

dpgalloway:
Was this ever implemented? How can we change logout from POST to Redirect?

cjbarth:
@dpgalloway , it doesn't appear so: https://github.com/node-saml/node-saml/blob/fd784087fc73697573d8389e508e83b42787d879/src/saml.ts#L1413

If you would like to take another crack at it, feel free. I'm not sure why @simonsmadsen decided to close this the day we provided feedback; maybe he'll comment on this PR and let us know.

dpgalloway:
@cjbarth Thanks for the quick reply. It's weird because the README mentions it but provides no actual instructions to implement (https://github.com/node-saml/passport-saml/blob/master/README.md#slo-single-logout):

> Passport-SAML has built in support for SLO including
> 
> Signature validation
> IdP initiated and SP initiated logouts
> Decryption of encrypted name identifiers in IdP initiated logout
> **Redirect and POST SAML Protocol Bindings**

I'll try and take a crack at it this week.

cjbarth:
Could have been an oversight when it was originally coded. I wasn't hear for all of that. Keep in mind that you'll probably want to add this to the `node-saml` project for a 4.0 release which is hopefully coming up soon.

PR #447, opened 2020-08-26
Return object for XML-valued AttributeValues

mans0954:
Closes: #245 

[saml-core-2.0-os](http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) Section 2.7.3.1.1 says
```
The <AttributeValue> element supplies the value of a specified SAML attribute. It is of the
xs:anyType type, which allows any well-formed XML to appear as the content of the element.
```

This PR modifies the `attrValueMapper` function to return an object representing the parsed XML unless the attribute is a simple string, in which case the string is returned.

A test is included which also serves as an example of how this functionality can be used in the specific case of the deprecated [eduPersonTargetedID](https://wiki.refeds.org/display/STAN/eduPerson+2020-01#eduPerson2020-01-eduPersonTargetedID) attribute.

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -1009,7 +1009,7 @@ SAML.prototype.processValidlySignedAssertion = function(xml, samlResponseXml, in
                                );
 
       var attrValueMapper = function(value) {
-        return typeof value === 'string' ? value : value._;
+        return value._ ? value._ : value;
       };
 
       if (attributes) {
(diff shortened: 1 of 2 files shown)


cjbarth:
I haven't traced all the code paths on this one, but it looks reasonable to me. Thank you for the test.

markstos:
Thanks for referencing the spec, too. I hope to catch up on passport-saml mail soon.

@cjbarth feel free to merge PRs that look good.

PR #451, opened 2020-09-07
Patch `strategy.js` for default values in `authenticate` method

puranjayjain:
There is an error in passport and loopback 4 integration wherein `samlFallback` of undefined is not a function comes up. This is done to patch it


markstos:
Can you provide an automated test which triggers the error?

What's the oldest version of Node.js that supports this syntax?

puranjayjain:
Sure @markstos

puranjayjain:
No longer working with loopback 4

PR #457, opened 2020-09-23
Provide default options value to authenticate function when not provided

### LINKED ISSUE #424: [BUG] authenticate
If options is undefined, the authenticate function send an error :
TypeError: Cannot read property 'samlFallback' of undefined

Expected : 
the 'login-request' must be setted.

pilare:
Fixes #424
https://github.com/node-saml/passport-saml/issues/424

Provided default options value to authenticate, when options are not provided.
Prevents error:
Cannot read property 'samlFallback' of undefined

### DIFF
--- a/lib/passport-saml/strategy.js
+++ b/lib/passport-saml/strategy.js
@@ -32,7 +32,7 @@ function Strategy (options, verify) {
 
 util.inherits(Strategy, passport.Strategy);
 
-Strategy.prototype.authenticate = function (req, options) {
+Strategy.prototype.authenticate = function (req, options = {name:'saml'}) {
   var self = this;
 
   options.samlFallback = options.samlFallback || 'login-request';


cjbarth:
Thank you for submitting the fix for this. I have just two comments:

1. Could you please include a test to validate this. I know it is simple, but we are working hard at increasing and improving our test coverage.
2. I would like to hear from some of the other maintainers on this one. It would seem that a constant defined as all the defaults we want to use, like is done in some other places in the code, might be a more appropriate solution than just defining an object in the function declaration as is done here.

pilare:
@cjbarth I'm still working on integration using passport-saml in my current project. This fix does help me with my issue, but I have to look into this more to assess what impact could it have in other situations. I will go back to this to add testing cases, if nobody will do it earlier than me. I'm not sure, if the name attribute is even needed there, as it is never used, so it can be empty object instead. Glad to hear input from others as well.

cjbarth:
@pilare We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

Issue #169, opened 2016-09-14
Host parameter is ignored


vasuems:
I have configured host, callbackUrl parameters along with path. Path value is considered. host and callbackUrl values are ignored. Instead of the configured host value, ip is used to generate AssertionConsumerServiceUrl. For example, my AssertionConsumerServiceUrl should be https://domainnmae/login/callback. Instead the AssertionConsumerServiceUrl is http:ip/login/callback. Because of this, post the authentication, the request fail to reach the call back function.

pdspicer:
host option is used for metadata generation. when generating authnRequests, the value from req.headers.host is used (protocol comes from req.protocol as well if options.protocol is not set). Are you potentially using a reverse proxy that might be terminating SSL and/or not setting the host header properly?

nudoru:
I'm having the same issue. My app is configured as shown below but the AssertionConsumerServiceUrl always goes back to http and my app fails the redirect.

path: '/login/callback', 
callbackUrl: 'https://domain.com/login/callback',
entryPoint: 'https://domain.com/idp/',
protocol: 'https',

mx781:
Indeed it seems that passing callbackUrl in options gets ignored for some reason. We resolved the issue by passing `protocol='https://'` instead of just "https".

theaccordance:
Is there a justification for using `req.headers.host` over the defined value for `this.options.host`?  

This is causing problems on my end, as my clients are making a call to our backend (app.server.com) to start the SAML request via an alias (app1.server.com).  Since the header contains the alias, that is in turn being passed to third-party SAML providers, which aren't aware of our aliases & don't have them included on their whitelists.  

IMO if we're setting a host value, that should be the value used by the host before it uses an implied value.

ahavriluk:
I agree! There is no reason to pull the host from req.headers.host if this.options.host is provided.

cjbarth:
I believe there may be more at play here than it first appears. I would expect in the case of @theaccordance use-case, that you'd have another federation or a redirection in order to handle the different domains. It seems that it would actually be undesirable to be able to configure things in such a way as to allow piggy-backing on another federation; it might even be considered a security issue (though I do acknowledge that using a `host` header does have its own implications for security).

As @pdspicer mentioned, `options.host` is for metadata generation. It follows that the metadata that is generated should match the requests that your server should be handling, so really `options.host` and `req.headers.host` should match. In cases where they don't, I would expect that condition to be exceptional and for the authentication to fail.

[...]

PR #460, opened 2020-09-28
fix host parameter is ignored

lucas-dolsan:
fix host parameter is ignored, mentioned in this issue: [#169](https://github.com/node-saml/passport-saml/issues/169)

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -103,13 +103,12 @@ SAML.prototype.getCallbackUrl = function (req) {
   if (this.options.callbackUrl) {
     return this.options.callbackUrl;
   } else {
-    var host;
-    if (req.headers) {
-      host = req.headers.host;
-    } else {
-      host = this.options.host;
+    if(this.options.host) {
+      return this.options.host + this.options.path;
     }
-    return this.getProtocol(req) + host + this.options.path;
+    if (req.headers) {
+      return this.getProtocol(req) + req.headers.host + this.options.path;
+    } 
   }
 };
 


cjbarth:
Thank you for your involvement in the maintenance of `passport-saml`. It would be helpful if some tests could be included to validate that this bug stays fixed.

Having said that, while this is my first time looking at this bug, it does seem that this fix could have some unintended side effects, or potential security implications. I'll express my concerns on the bug thread so that all can comment.

markstos:
@cjbarth this seems like safer behavior than the old behavior. Befor we actually preferred trusting to use the incoming `host` header over the configured value. Considering that the incoming `host` header could be malicious, that seems less than ideal.

This PR reverses the logic, so the configured `host` header is preferred over the request `host` header.

My question is: _do we need to support req.headers.host at all?_ The safest option of all would be to not use the incoming host header at all, unless the spec requires it. 

@lucas-dolsan To help move this along, it would be helpful if you reference any SAML spec sections that relate to this behavior change.

cjbarth:
Unless I'm missing something, how would we properly support SLO or multi-federation if we don't trust the incoming `host` header, at least to compare it to something?

markstos:
@cjbarth Accepting it and comparing it to options.host would be even stricter than ignoring it. I'm on board with that.

srd90:
@markstos @cjbarth At least some IdP implementations verify among other things that ACS location provided in auth request match with one of the (possibly multiple) ACS locations listed at provisioned SP metadata (which can be manually crafted or automatically generated etc.):

>For the protection of the user, the valid Assertion Consumer Service location(s) associated with an SP are generally registered in metadata so that they can be checked by the IdP. This helps to prevent the delivery of personal information to unauthorized parties.

source: https://wiki.shibboleth.net/confluence/display/CONCEPT/AssertionConsumerService ( quote was from this version https://wiki.shibboleth.net/confluence/pages/viewpage.action?pageId=34472059 )

And
>SPs support SSO protocols by including one or more `<md:AssertionConsumerService>` endpoint elements in their metadata. These are the locations to which the IdP will eventually send the user at the SP. By enumerating them in the metadata, the IdP can ensure that the user's information is sent only to authorized locations.

source: https://wiki.shibboleth.net/confluence/display/CONCEPT/MetadataForSP#MetadataForSP-AssertionConsumerServices (exact version of page https://wiki.shibboleth.net/confluence/pages/viewpage.action?pageId=4358454#MetadataForSP-AssertionConsumerServices )

srd90:
btw isn't it so that if `options.host` is not provided or value is set to `undefined/falsy` then `passport-saml`'s initialization code shall always set value to `localhost`.
https://github.com/node-saml/passport-saml/blob/43465d6b4957218b05ca040dd5c8d853a1208cbd/lib/passport-saml/saml.js#L35-L37
I.e. `options.host` is never `falsy`

Current "callback url resolving algorithm" is
https://github.com/node-saml/passport-saml/blob/43465d6b4957218b05ca040dd5c8d853a1208cbd/lib/passport-saml/saml.js#L101-L114
This PR's **current** (at the time of writing this comment) version looks like this:
```javascript
 SAML.prototype.getCallbackUrl = function (req) { 
     // Post-auth destination 
   if (this.options.callbackUrl) { 
     return this.options.callbackUrl; 
   } else { 
     if(this.options.host) { 
       return this.options.host + this.options.path; 
     } 
     if (req.headers) { 
       return this.getProtocol(req) + req.headers.host + this.options.path; 
     }  
   } 
 }; 
```
so - regardless of what anyone might think about using `req.host` for anything - this PR's actual result would not be "reverting the logic (i.e. preferring `options.host` over `req.host`)" because `req.host` would be always ignored because it is in the "unreachable execution path".
@markstos @cjbarth

cjbarth:
Good sleuthing @srd90 ! I believe your assessment is right on. In that case, we definitely don't want to land this PR as-is.

I feel like we really need a test suite put in place for this block of code to make sure that current desired behavior doesn't break and includes a test that is broken without some fix. Given that many people are using multiple federations with proper callback behavior, and there are "workarounds" for the reported "incorrect" behavior, I'm inclined to make sure we really don't introduce a regression here by being extra careful here.

Additionally, I'm not convinced that the usage that is exhibiting this `passport-saml` behavior is actually correct usage (see my comments in #169). I'd like to get a clearer idea of the usage, problem, and potential solution worked out over there before we move too much farther with a PR to address #169.

@lucas-dolsan Are you in a position to help us be a little more thorough with this issue by including a test suite?

cjbarth:
@lucas-dolsan We will be releasing v3.x in the upcoming weeks. If you would like this included in that release, please address the latest comments and we'll get this landed.

cjbarth:

PR #461, opened 2020-10-06
fix: add catch block to NameID decryption

bryan-lockhart:
the callback invocation was being missed in case of errors during the NameID decryption


cjbarth (approved):
Thanks for the tests!

markstos:
Looks the right way to bridge promises and callbacks to me.

PR #463, opened 2020-10-12
Add GitHub Actions as Continuos Integration provider

walokra:
Travis CI had been removed when project was renamed to node-saml and thus tests are not run anymore to verify package state. This PR replaces Travis CI with GitHub Actions as it does the job and keeps everything in one place.

The CI task runs lint and tests on Node.js 10 and 12. Build Status is shown in a badge.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x]
+        node-version: [10.x, 12.x, 14.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Build Status
 
 on: [push]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,24 @@
...
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
```
cjbarth, line 12: Node 14 is released, but hasn't entered LTS, I think we should test against that too.

cjbarth (changes requested):

.github/workflows/workflow.yml
walokra: Node 14 LTS starts 2020-10-27 so maybe it's reasonable to already enable testing againts it although the documentation states that "We only support Long-Term Support versions of Node.".

cjbarth: I feel like, while we only support LTS versions, if a version is Current and will become LTS, we should probably add it to tests so that we aren't surprised by something.

cjbarth (approved):

PR #464, opened 2020-10-13
Only make an attribute an object if it has child elements

### LINKED ISSUE #459: [BUG] Regression between 1.2.0 and 1.3.5 for undefined values
When the IdP sends the following attribute description:
```
            <saml2:Attribute Name="attributeName"
                             NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified"
                             >
                <saml2:AttributeValue xmlns:xs="http://www.w3.org/2001/XMLSchema"
                                      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                                      xsi:type="xs:string"
                                      />
            </saml2:Attribute>
```
With passport-saml 1.2.0 I get the value: `undefined` for `profile.attributeName`, as I expect. But with passport-saml 1.3.5, I get this object instead:
```
{
    '$': {
      'xmlns:xs': 'http://www.w3.org/2001/XMLSchema',
      'xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance',
      'xsi:type': 'xs:string'
    }
  }
```

Was that change intended, or is it a bug ? It sure broke my application when I updated passport-saml to avoid a security issue, as I did not expect to get something like that...

mans0954:
This PR

* Adds a test for an empty SAML AttributeValue, which should be represented as `undefined`
* Changes the `attrValueMapper` behaviour so that it returns the value of `_` (which will be the char value string or `undefined`) unless the SAML AttributeValue has child elements, in which case it returns the AttributeValue as an object.

Branch https://github.com/node-saml/passport-saml/tree/csh-issue-459-attr-value-regression-ec presents an alternative approach which would have wider impact/implications.

Closes #459

### DIFF
--- a/lib/passport-saml/saml.js
+++ b/lib/passport-saml/saml.js
@@ -1009,7 +1009,8 @@ SAML.prototype.processValidlySignedAssertion = function(xml, samlResponseXml, in
                                );
 
       var attrValueMapper = function(value) {
-        return value._ ? value._ : value;
+        const hasChildren = Object.keys(value).some((cur)=> { return (cur!=='_' && cur!=='$'); });
+        return (hasChildren) ? value : value._;
       };
 
       if (attributes) {
--- a/test/tests.js
+++ b/test/tests.js
@@ -1384,7 +1384,29 @@ describe( 'passport-saml /', function() {
         });
       });
 
-
+      it( 'An undefined value given with an object should still be undefined', function( done ) {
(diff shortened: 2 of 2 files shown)


cjbarth:
This seems like the correct solution to resolve the regression, however, I am interested in hearing comments from others about the bigger change for the next semver major release.

Issue #456, opened 2020-09-21
"invalid signature" after updating passport saml


yj188:
passport saml version: 1.1.0 --> 1.3.5
xml-crypto: 1.4.0 --> 1.5.3
xml-encryption: 0.11.2 ->1.2.1

Error: Invalid signature
    at /var/task/node_modules/passport-saml/lib/passport-saml/saml.js:611:15
    at _fulfilled (/var/task/node_modules/q/q.js:854:54)
    at /var/task/node_modules/q/q.js:883:30
    at Promise.promise.promiseDispatch (/var/task/node_modules/q/q.js:816:13)
    at /var/task/node_modules/q/q.js:624:44
    at runSingle (/var/task/node_modules/q/q.js:137:13)
    at flush (/var/task/node_modules/q/q.js:125:13)
    at processTicksAndRejections (internal/process/task_queues.js:79:11)

yj188:
Update xml-crypto back to 1.4.0 resolve the issue

cjbarth:
Was a corresponding bug filed with `xml-crypto` to get this resolved? I hate to just leave us stuck on v1.4.0.

markstos:
@cjbarth There are newer releases of xml-crypto out. Could test an upgrade with them? 1.5.3 or 2.0.0. The 2.0.0 release addresses some security concerns but is a breaking change.

cjbarth:
@yj188 Could you please provide a reproduction for the error you encountered? Ideally, it would be nice to have a test added that would pass with v1.4.0 and fail with v1.5.3.

Harsha3878:
passport saml version: 1.3.4 --> 1.3.5
xml-crypto: 1.4.0 
xml-encryption: 1.2.1

Error: Invalid signature
   \node_modules\passport-saml\lib\passport-saml\saml.js:611:15
    node_modules\q\q.js:854:54)
    node_modules\q\q.js:883:30
    at Promise.promise.promiseDispatch (\node_modules\q\q.js:816:13)
    at node_modules\q\q.js:624:44
    at runSingle (\node_modules\q\q.js:137:13)
    at flush (\node_modules\q\q.js:125:13)
    at processTicksAndRejections (internal/process/task_queues.js:79:11)

cjbarth:
@Harsha3878 , are you saying that v1.3.4 works, but v1.3.5 fails with the current test suite?

Harsha3878:
[...]

PR #465, opened 2020-10-15
Upgrade xml-crypto dependancy

mans0954:
* Bump xml-crypto from 1.4.0 to 1.5.3
* Include package-lock.json in repo

May possibly help with https://github.com/node-saml/passport-saml/issues/456 ?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2692 @@
+{
+  "name": "passport-saml",
+  "version": "1.3.5",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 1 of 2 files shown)


cjbarth:
@mans0954 , how would this help with #456? It seems that #465 is reporting that moving to v1.5.3 is what caused the problem in the first place. Unfortunately neither @yj188 nor @Harsha3878 provided a clear reproduction or a test that can be used to figure out what is going on.

mans0954:
> @mans0954 , how would this help with #456? It seems that #465 is reporting that moving to v1.5.3 is what caused the problem in the first place. Unfortunately neither @yj188 nor @Harsha3878 provided a clear reproduction or a test that can be used to figure out what is going on.

The test pass with v1.5.3 when I run them - adding `package-lock.json` should ensure that we're all using the same sub-dependencies. I have no specific reason to think that users having different sub-dependencies installed is the issue in #465 - it just eliminates one possibility.

cjbarth (approved):

PR #466, opened 2020-10-21
Update README.md

mohankoppuravuri:
We are getting the error "Cannot read value stripPrefix of undefined".
Passport JS is not converting xml into JS (where we get the profile object).
We have depend on extrernal module to do this work. So I am adding this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Passport-SAML has been tested to work with Onelogin, Okta, Shibboleth, [SimpleSA
 ## Installation
 
     $ npm install passport-saml
+    $ npm i xml2js
 
 ## Usage
 


cjbarth:
This shouldn't be needed. The package you reference here is already listed as a dependency in [package.json](https://github.com/node-saml/passport-saml/blob/026edf2a422f70ee8f483c902594bf29a27def3c/package.json#L35).

mohankoppuravuri:
Yeah you are correct!

But don't know the reason, My code was working only after installing the module separately.

cjbarth:
This seems like it might be more your environment than a problem with node/npm. In any case, if a dependency isn't being installed, then it isn't a problem with `passport-saml`, this would need to be addressed with the npm folks.

PR #472, opened 2020-10-28
Pass 'options' to generateLogoutRequest so that it contains vetuma-extension element

jajvirta:
Mainline passport-saml has dropped the 'options' from the generateLogoutRequest function call, which means that options.vetumaLang gets dropped and the logoutrequest no longer has vetuma-extension. This in turn means that Single Logout page won't be localized correctly.


jajvirta:
Wrong project, sorry!

PR #477, opened 2020-10-29
Primary files use typescript

gugu:
This PR switches to valid TS other files (except tests)

SAML module now uses native classes, Strategy still uses a functional approach with .prototype because of the dodgy way of override in MultiSAML

Lint returns 0 errors now, but typescript is not in a strict mode


markstos:
Like the other refactor, I think this one can be trusted as all tests contiue to pass.

PR #478, opened 2020-10-29
outdated Q library was removed

gugu:
Q functions were replaced with corresponding native alternatives,
switching to async/await is not part of this PR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "passport-saml",
-  "version": "1.3.5",
+  "version": "1.4.2",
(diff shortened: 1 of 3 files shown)


cjbarth (approved):
For the sake of the reviewer, it would be nice if we could have a PR to clear up as many of these warnings from our GitHub Actions as possible :)

Otherwise, thanks for modernizing things so much. This will no doubt help many consumers of this library.

markstos:
QA Log
======

  * Reviewed diff, looks good to me.  I like that unlike an earlier attempt at this, no shim functions are used, just native functions.

Issue #482, opened 2020-10-30
[BUG] Type errors


robcresswell:
```
> tsc --project tsconfig.json

node_modules/passport-saml/lib/passport-saml/index.d.ts:3:8 - error TS1259: Module '"/home/robcresswell/snyk/my-service/node_modules/passport-saml/lib/passport-saml/strategy"' can only be default-imported using the 'esModuleInterop' flag

3 import Strategy from './strategy';
         ~~~~~~~~

  node_modules/passport-saml/lib/passport-saml/strategy.d.ts:2:1
    2 export = _default;
      ~~~~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

node_modules/passport-saml/lib/passport-saml/saml.d.ts:2:8 - error TS1192: Module '"crypto"' has no default export.

2 import crypto, { KeyLike } from 'crypto';
         ~~~~~~

node_modules/passport-saml/lib/passport-saml/types.d.ts:1:13 - error TS1259: Module '"/home/robcresswell/snyk/my-service/node_modules/@types/express/index"' can only be default-imported using the 'esModuleInterop' flag

1 import type express from 'express';
              ~~~~~~~

  node_modules/@types/express/index.d.ts:108:1
    108 export = e;
        ~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

node_modules/passport-saml/lib/passport-saml/types.d.ts:2:8 - error TS1259: Module '"/home/robcresswell/snyk/my-service/node_modules/@types/passport/index"' can only be default-imported using the 'esModuleInterop' flag

2 import passport from 'passport';
         ~~~~~~~~

  node_modules/@types/passport/index.d.ts:169:1
    169 export = passport;
        ~~~~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

This seems to boil down to "please don't build a library with `esModuleInterop: true` because then all of your consumers also need to do this", but maybe I'm misreading the issue.

```
passport: 0.4.1
passport-saml: 1.5.0
```

Note if anyone else is blocked by this, the "least-hacky" workaround is probably to enable `skipLibCheck: true` or downgrade.

PR #483, opened 2020-10-30
fix: disable esmoduleInterop setting

robcresswell:
This patch disables the `esmoduleInterop` setting that causes type issues when the library is used by applications that do not transform their code in this way.

Note the workaround for the `strategy.ts` file (import = require()) is not ideal, but the export in that file is not a valid ES export, so using this TS workaround to get past it.

fixes: #482


gugu (approved):

gugu:
I agree, personally, I don't like esmoduleInterop setting, I forgot to turn it off

PR #484, opened 2020-10-30
TS strict types

gugu:
This PR changes a lot - typescript works with "strict": true

### DIFF
--- a/src/passport-saml/saml.ts
+++ b/src/passport-saml/saml.ts
@@ -1359,7 +1359,7 @@ class SAML {
 
     const wrappedKey = [
       '-----BEGIN PRIVATE KEY-----',
-      ...key.match(/.{1,64}/g)!,
+      ...(key.match(/.{1,64}/g) ?? []),
       '-----END PRIVATE KEY-----',
       ''
     ].join('\n');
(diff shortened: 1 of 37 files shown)


src/passport-saml/saml.ts
```diff
@@ -1333,7 +1359,7 @@ class SAML {
 
     const wrappedKey = [
       '-----BEGIN PRIVATE KEY-----',
-      ...key.match(/.{1,64}/g),
+      ...key.match(/.{1,64}/g)!,
```
markstos, line 1362: I looked up what the "!" was for here. It expresses that the value cannot be `null` or `undefined`. 

https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

markstos:
QA Log
=======

  * Reviewed diff
  * Noticed test coverage remains intact.

markstos:
@gugu is the "draft" status of this PR correct?

src/passport-saml/saml.ts
gugu: yes, JS had a problem - it did not check if key.match(/.{1,64}/) actually finds something and returns a match. Current PR copies this behavior. Here is transpiled version:

image

Fixed in 36263b2

gugu:
Opened #485 instead

PR #485, opened 2020-10-31
strict TS types, Strategy and MultiSamlStrategy use native classes

gugu:
This is the change worth of major version change

### DIFF
--- a/src/passport-saml/saml.ts
+++ b/src/passport-saml/saml.ts
@@ -16,16 +16,8 @@ import * as algorithms from './algorithms';
 import { signAuthnRequestPost } from './saml-post-signing';
 import type { Request } from 'express';
 import { ParsedQs } from 'qs';
-import { AuthenticateOptions, AuthorizeOptions, CertCallback, Profile, RequestWithUser } from './types';
+import { AuthenticateOptions, AuthorizeOptions, CertCallback, Profile, RequestWithUser, SamlScopingConfig } from './types';
 const { xpath } = xmlCrypto;
-
-interface SAMLUser {
-    sessionIndex: any;
-    spNameQualifier: any;
-    nameIDFormat: any;
-    nameID: any;
-    nameQualifier: null;
-}
 
 interface NameID {
     value: string | null;
@@ -33,7 +25,7 @@ interface NameID {
 }
 import { SamlIDPEntryConfig, SamlIDPListConfig } from './types';
 
-function processValidlySignedPostRequest(self: SAML, doc: any, dom: Document, callback: (err: Error | null, profile?: Profile | undefined, loggedOut?: boolean | undefined) => void) {
+function processValidlySignedPostRequest(self: SAML, doc: any, dom: Document, callback: (err: Error | null, profile?: Profile, loggedOut?: boolean) => void) {
   const request = doc.LogoutRequest;
   if (request) {
     const profile = {} as Profile;
@@ -95,6 +87,7 @@ function callBackWithNameID(nameid: Node, callback: (err: Error | null, value: N
 }
 
 export interface SAMLOptions {
+  scoping: SamlScopingConfig;
   xmlSignatureTransforms: string[];
   digestAlgorithm: string;
   providerName: string;
@@ -333,7 +326,7 @@ class SAML {
       }
 
       if (this.options.scoping) {
-        const scoping = {
+        const scoping: Record<string, any> = {
           '@xmlns:samlp': 'urn:oasis:names:tc:SAML:2.0:protocol',
         };
 
@@ -343,13 +336,13 @@ class SAML {
 
         if (this.options.scoping.idpList) {
           scoping['samlp:IDPList'] = this.options.scoping.idpList.map((idpListItem: SamlIDPListConfig) => {
-            const formattedIdpListItem = {
+            const formattedIdpListItem: Record<string, any> = {
               '@xmlns:samlp': 'urn:oasis:names:tc:SAML:2.0:protocol',
             };
 
             if (idpListItem.entries) {
               formattedIdpListItem['samlp:IDPEntry'] = idpListItem.entries.map((entry: SamlIDPEntryConfig) => {
-                const formattedEntry = {
+                const formattedEntry: Record<string, any> = {
                   '@xmlns:samlp': 'urn:oasis:names:tc:SAML:2.0:protocol',
                 };
 
--- a/src/passport-saml/saml.ts
+++ b/src/passport-saml/saml.ts
@@ -8,17 +8,34 @@ import * as xmldom from 'xmldom';
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as xmlbuilder from 'xmlbuilder';
+// @ts-expect-error no typings for xml-encryption
 import * as xmlenc from 'xml-encryption';
 import * as util from 'util';
 import {CacheProvider as InMemoryCacheProvider} from './inmemory-cache-provider';
 import * as algorithms from './algorithms';
 import { signAuthnRequestPost } from './saml-post-signing';
 import type { Request } from 'express';
+import { ParsedQs } from 'qs';
+import { AuthenticateOptions, AuthorizeOptions, CertCallback, Profile, RequestWithUser } from './types';
+const { xpath } = xmlCrypto;
 
-function processValidlySignedPostRequest(self: SAML, doc, dom, callback) {
+interface SAMLUser {
+    sessionIndex: any;
+    spNameQualifier: any;
+    nameIDFormat: any;
+    nameID: any;
+    nameQualifier: null;
+}
+
+interface NameID {
+    value: string | null;
+    format: string | null;
+}
+
+function processValidlySignedPostRequest(self: SAML, doc: any, dom: Document, callback: (err: Error | null, profile?: Profile | undefined, loggedOut?: boolean | undefined) => void) {
   const request = doc.LogoutRequest;
   if (request) {
-    const profile: Record<string, any> = {};
+    const profile = {} as Profile;
     if (request.$.ID) {
         profile.ID = request.$.ID;
     } else {
@@ -55,7 +72,7 @@ function processValidlySignedPostRequest(self: SAML, doc, dom, callback) {
 }
 
 
-function processValidlySignedSamlLogout(self: SAML, doc, dom, callback) {
+function processValidlySignedSamlLogout(self: SAML, doc: any, dom: Document, callback: (err: Error | null, profile?: Profile | null | undefined, loggedOut?: boolean | undefined) => void) {
   const response = doc.LogoutResponse;
   const request = doc.LogoutRequest;
 
@@ -68,7 +85,7 @@ function processValidlySignedSamlLogout(self: SAML, doc, dom, callback) {
   }
 }
 
-function callBackWithNameID(nameid, callback) {
+function callBackWithNameID(nameid: Node, callback: (err: Error | null, value: NameID) => void) {
   const format = xmlCrypto.xpath(nameid, "@Format") as Node[];
   return callback(null, {
     value: nameid.textContent,
@@ -76,7 +93,15 @@ function callBackWithNameID(nameid, callback) {
   });
 }
 
-interface SAMLOptions {
+export interface SAMLOptions {
+  xmlSignatureTransforms: string[];
+  digestAlgorithm: string;
+  providerName: string;
+  attributeConsumingServiceIndex: string | null;
+  RACComparison: any;
+  authnContext: string | string[];
+  disableRequestedAuthnContext: boolean;
+  disableRequestACSUrl: boolean;
   acceptedClockSkewMs: number;
   protocol: string;
   host: string;
@@ -94,13 +119,14 @@ interface SAMLOptions {
   cacheProvider: InMemoryCacheProvider;
   issuer: string;
   identifierFormat: string;
-  cert: string | string[];
+  cert: string | string[] | CertCallback;
   passive: boolean;
   decryptionPvk: string;
   logoutCallbackUrl: string;
   validateInResponseTo: boolean;
   requestIdExpirationPeriodMs: number;
   audience: string;
+  forceAuthn: boolean;
 }
 
 
@@ -108,11 +134,11 @@ class SAML {
   options: SAMLOptions;
   cacheProvider: InMemoryCacheProvider;
 
-  constructor(options) {
+  constructor(options: Partial<SAMLOptions>) {
     this.options = this.initialize(options);
     this.cacheProvider = this.options.cacheProvider;
   }
-  initialize(options) {
+  initialize(options: Partial<SAMLOptions>): SAMLOptions {
     if (!options) {
       options = {};
     }
@@ -184,14 +210,14 @@ class SAML {
       options.RACComparison = 'exact';
     }
 
-    return options;
+    return options as SAMLOptions;
   }
 
-  getProtocol(req) {
+  getProtocol(req: Request | {headers?: undefined, protocol?: undefined}) {
     return this.options.protocol || (req.protocol || 'http').concat('://');
   }
 
-  getCallbackUrl(req) {
+  getCallbackUrl(req: Request | {headers?: undefined, protocol?: undefined}) {
     // Post-auth destination
     if (this.options.callbackUrl) {
       return this.options.callbackUrl;
@@ -214,7 +240,7 @@ class SAML {
     return new Date().toISOString();
   }
 
-  signRequest(samlMessage) {
+  signRequest(samlMessage: Record<string, any>) {
     const samlMessageToSign: Record<string, any> = {};
     samlMessage.SigAlg = algorithms.getSigningAlgorithm(this.options.signatureAlgorithm);
     const signer = algorithms.getSigner(this.options.signatureAlgorithm);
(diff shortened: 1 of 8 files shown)


cjbarth:
@gugu Can we do something to clear up these GitHub action warnings? It makes the code harder to review with so many warnings spreading out the code. Also, since this is a major change, it seems we have some time to clear up these kinds of things.

gugu:
I agree with you about the warning. It is possible to disable warnings in specific file for this PR:
image

And I'll provide a special PR for fixing warnings

markstos (approved):
QA Log
======

  * Reviewed diff. Noted all tests still pass and all changes are related to typing.

markstos:
@gugu Looks good. Could you address the `saml.ts` conflict?

src/passport-saml/saml.ts
```diff
@@ -459,7 +485,7 @@ class SAML {
       additionalParams[k] = optionsAdditionalParams[k];
     });
 
-    let optionsAdditionalParamsForThisOperation = {};
+    let optionsAdditionalParamsForThisOperation: Record<string, string> = {};
```
cjbarth, line 488: I've noticed that you use this `Record<>` construct a lot in this very generic way. I'm not used to seeing it used this way. Instead, I've more accustomed to seeing `{[K: string]: string}`. Is there a reason this notation was chosen? It seems much more C#-y than JavaScript-y.

src/passport-saml/saml.ts
```diff
@@ -473,14 +499,14 @@ class SAML {
 
     overrideParams = overrideParams || {};
     Object.keys(overrideParams).forEach(function(k) {
-      additionalParams[k] = overrideParams[k];
+      additionalParams[k] = overrideParams![k];
```
cjbarth, line 502: I'm surprised that you need the `!` here at all. I expect that TypeScript will note that all values of `k` exist because of the iterator. What am I missing?

src/passport-saml/saml.ts
```diff
@@ -503,19 +529,19 @@ class SAML {
...
       if (err) {
         return callback(err);
       }
 
       const operation = 'authorize';
       const additionalParameters = this.getAdditionalParams(req, operation);
-      const samlMessage = {
-        SAMLRequest: buffer.toString('base64')
+      const samlMessage: Record<string, string | number> = {
+        SAMLRequest: buffer!.toString('base64')
```
cjbarth, line 544: How can we assert this will be non-`null`?

src/passport-saml/saml.ts
```diff
@@ -547,21 +573,21 @@ class SAML {
...
 
-    this.generateAuthorizeRequest(req, this.options.passive, true, (err, request) => {
+    this.generateAuthorizeRequest(req, this.options.passive, true, (err: Error | null, request?: string) => {
       if (err) {
         return callback(err);
       }
 
       if (this.options.skipRequestCompression) {
-        getAuthorizeFormHelper(null, Buffer.from(request, 'utf8'));
+        getAuthorizeFormHelper(null, Buffer.from(request!, 'utf8'));
```
cjbarth, line 582: I feel like I'm missing some line of reasoning here in your code. I see lots of `!` being used when I expect `?` to be used. We need to check if these values are `null`, not assert that they aren't. Can you please comment on what I'm missing?

I won't flag the rest of my questions of this nature at this time.

src/passport-saml/saml.ts
```diff
@@ -628,24 +654,24 @@ class SAML {
 
     const signature = signatures[0];
```
cjbarth, line 655: `signatures` should be defined above on 648 instead of here or on line 657 using `as string`

src/passport-saml/saml.ts
```diff
@@ -628,24 +654,24 @@ class SAML {
...
     });
   }
 
 // This function checks that the |signature| is signed with a given |cert|.
-  validateSignatureForCert(signature, cert, fullXml, currentNode) {
+  validateSignatureForCert(signature: string | Node, cert: string, fullXml: string, currentNode: HTMLElement) {
     const sig = new xmlCrypto.SignedXml();
     sig.keyInfoProvider = {
-      file: undefined,
+      file: '',
```
cjbarth, line 665: Is this change really correct? Shouldn't it be `null` or simply not defined?

src/passport-saml/saml.ts
```diff
@@ -673,10 +699,10 @@ class SAML {
       if (!Object.prototype.hasOwnProperty.call(doc, 'documentElement'))
         throw new Error('SAMLResponse is not valid base64-encoded XML');
 
-      inResponseTo = xmlCrypto.xpath(doc, "/*[local-name()='Response']/@InResponseTo");
+      const inResponseToNodes = xmlCrypto.xpath(doc, "/*[local-name()='Response']/@InResponseTo") as Node[];
```
cjbarth, line 702: Why aren't the `xmlCrypto` types handling this case correctly?

src/passport-saml/saml.ts
```diff
@@ -716,18 +742,18 @@ class SAML {
 
         const xmlencOptions = { key: this.options.decryptionPvk };
         return util.promisify(xmlenc.decrypt).bind(xmlenc)(encryptedAssertionXml, xmlencOptions)
-        .then(decryptedXml => {
+        .then((decryptedXml: string) => {
           const decryptedDoc = new xmldom.DOMParser().parseFromString(decryptedXml);
-          const decryptedAssertions = xmlCrypto.xpath(decryptedDoc, "/*[local-name()='Assertion']");
+          const decryptedAssertions = xmlCrypto.xpath(decryptedDoc, "/*[local-name()='Assertion']") as HTMLElement[];
```
cjbarth, line 747: Again, it seems we might be doing something sub-optimal or the types for `xmlCrypto` need to be updated.

src/passport-saml/saml.ts
```diff
@@ -877,9 +903,9 @@ class SAML {
 
       return this.certsToCheck()
         .then(certs => {
-          const hasValidQuerySignature = certs.some(cert => {
+          const hasValidQuerySignature = certs!.some(cert => {
             return this.validateSignatureForRedirect(
-              urlString, container.Signature, container.SigAlg, cert
+              urlString, container.Signature as string, container.SigAlg as string, cert
```
cjbarth, line 908: These things feel more like a hope than an assurance. Bad incoming data would cause this not to be true.

Overall, I'm looking at adding these types, not as a way to make TypeScript happy, but to actually express the truth. I'd rather leave gaps in the types where we don't know for sure, than to express something that may not be true just because it should be.

I won't comment on any more of these `as type` at this time.

src/passport-saml/saml.ts
```diff
@@ -955,20 +981,20 @@ class SAML {
...
       explicitRoot: true,
       explicitCharkey: true,
       tagNameProcessors: [xml2js.processors.stripPrefix]
     };
     const nowMs = new Date().getTime();
-    const profile: Record<string, any> = {};
-    let assertion;
-    let parsedAssertion;
+    const profile = {} as Profile;
+    let assertion: any;
```
cjbarth, line 993: I really don't like using `any`. Perhaps a placeholder that is a very generic type could be used here instead of `any`. That still allows flexibility, but also shows that we've though about what is actually happening here. Really, these are objects coming back, so at least we know that much We can extend them later with properties we know as well and then include a generic `[K: string]: any` or something like that. That will help with IntelliSense.

src/passport-saml/strategy.ts
```diff
@@ -1,131 +1,131 @@
...
-  this._saml = new saml.SAML(options);
-  this._passReqToCallback = !!options.passReqToCallback;
-  this._authnRequestBinding = options.authnRequestBinding || 'HTTP-Redirect';
-}
+    // Customizing the name can be useful to support multiple SAML configurations at the same time.
+    // Unlike other options, this one gets deleted instead of passed along.
+    if  (options.name) {
+      this.name  = options.name;
+    }
+    else {
```
cjbarth, line 30: We use different style `else` in this project, even in this single file. Can we adjust our linter to catch/fix this? I personally like `} else {`, but as long as it is consistent.

tsconfig.json
```diff
@@ -25,7 +25,7 @@
     // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
 
     /* Strict Type-Checking Options */
-    "strict": false,                           /* Enable all strict type-checking options. */
+    "strict": true,                           /* Enable all strict type-checking options. */
```
cjbarth, line 28: I should have reviewed this file first :) I would say that it might be wise for sanity purposes to turn these checks on one-at-a-time so that the PRs are smaller. As you can see I made a lot of comments about all sorts of these different things. I would like a second set of eyes on a PR this large, and, given how much we've broken others peoples code with our TypeScript changes, it seems prudent to bite off smaller chunks at a time.

I'm all for starting our 2.0 release cycle and not publishing until we are done with all these changes. If there is a bug-fix or security thing that needs to be addressed, I think we can branch from the current 1.5 tag and fix it there and then merge that back in to `master`.

Thoughts?

cjbarth (changes requested):
I worry about this "big gun" approach to adding types. I feel like types were added in places just to make TypeScript happy, even if we aren't actually sure of the types we are using.

src/passport-saml/saml.ts
```diff
@@ -8,17 +8,34 @@ import * as xmldom from 'xmldom';
 import * as url from 'url';
 import * as querystring from 'querystring';
 import * as xmlbuilder from 'xmlbuilder';
+// @ts-expect-error no typings for xml-encryption
```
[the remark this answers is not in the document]
markstos: `xml-encryption` is now managed by `auth0`. They may be interested in maintaining TypeScript types in the main repo. https://github.com/auth0/node-xml-encryption

src/passport-saml/strategy.ts
markstos: My preference would to enforce code formatting with `eslint` or `prettier`. 

I like the idea of using `prettier` for code formatting and `eslint for code logic conventions`.

Automating code format checking can be especially valuable for open source projects that receive a lot of pull requests from new contributors who are not familiar with the conventions of the project.  Improving code formatting automation can be a separate issue and PR, though!

markstos:
@cjbarth I appreciate the thorough peer review. I am comfortable with the scope of this PR because it focuses narrowly on typing even though there a fair amount of typing changes throughout the project. You pointed out some places where maybe we could be even stricter in our typing, which could be beneficial.

In terms of release management, I'd like to focus on getting a 2.0 release in the next day or two. It would nice to avoid duplicating effort with a 1.5.x release if we can avoid it.

Although there are several places where this pull request could be even stricter, all tests already pass and further improving our strictness feels like something we could do in a point release. It should not break anyone's code who is usual the module as intended.

cjbarth:
@markstos I guess what I'm getting at is that we can make all the typing changes that are accurate, and just not turn on the TypeScript setting to enforce all the strict types. That will allow progressive improvement in ways that it really matters.

Speaking from experience, putting in the wrong type, or `any`, is just a bad idea. It tends to live forever. It also tends to misinform future readers of the code that certain things are true, when, in fact, those types were just put in once-upon-a-time to meet some arbitrary `strict` checking.

So, we could just remove all the `!` assertions that a variable will never be `null`, which I think in every case wasn't true. We can also remove all the `as type` unless we are sure it is accurate, which I don't think any actually are. We can also remove all the `any` until we actually create a valid type for those variables. Since the code passes TypeScript compilation now, it will continue to do so if compilation settings aren't changed.

@gugu, what are your thoughts on this?

gugu:
I'll go through comments tomorrow, I agree about `!` marks and will address these issues

markstos:
@cjbarth I'm aligned with your feedback. To get the benefits of TypeScript we need to type as strictly as we can and now is the best time to do that. 

@gugu Thanks for the work and reviewing @cjbarth feedback!

cjbarth:
> we need to type as strictly as we can

...but no stricter, which is what it feels like were doing here.

@gugu has done amazing work and I don't want to seem unappreciative, I really am as TypeScript is a wonderful tool. I just feel we're using the tool incorrectly; a hammer can drive a screw and it works if your goal is to have all the fasteners flush with the surface, but the screw won't do its job as well as if a screwdriver was used.


> now is the best time to do that

TypeScript is a progressive enhancement technology, so "now" done the wrong way actually increases harm, with no advantage, that "when it is ready" avoids.

I won't block this landing, but I can't approve it like this either. I will keep a close eye on this to provide feedback. I want types, I really do, but only the right types. If we can't get them right now, we can totally leave them off until we can and we loose nothing.

src/passport-saml/saml.ts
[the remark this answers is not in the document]
gugu: I think it is out of the scope of this PR. I use // @ts-expect-error instead of @ts-ignore because compilation will fail when node-xml-encryption adds typings. Fixing this problem will take unknown period of time, so I suggest keeping this comment for now

gugu:
I've added stricter types. Added XMLInput/XMLOutput types and types for Metadata. I did not alter `!` marks for now

Let's keep a balance between strict types and readability of the code, that's the awesome part of TypeScript. Sometimes it is better to add `!` or force type casting instead of writing code just to make typescript happy

src/passport-saml/saml.ts
gugu: no, xmlCrypto types are correct, they don't have any insights about input value (does it select an attribute or an element) and we adding restrictions is our responsibility

src/passport-saml/saml.ts
```diff
@@ -8,17 +8,34 @@ import * as xmldom from 'xmldom';
...
+    nameQualifier: null;
+}
+
+interface NameID {
+    value: string | null;
+    format: string | null;
+}
 
-function processValidlySignedPostRequest(self: SAML, doc, dom, callback) {
+function processValidlySignedPostRequest(self: SAML, doc: any, dom: Document, callback: (err: Error | null, profile?: Profile | undefined, loggedOut?: boolean | undefined) => void) {
```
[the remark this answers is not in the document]
gugu: done

src/passport-saml/saml.ts
```diff
@@ -247,7 +273,7 @@ class SAML {
       }
     })()
     .then(() => {
-      const request = {
+      const request: any = {
```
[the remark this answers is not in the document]
gugu: done

src/passport-saml/saml.ts
gugu: keyInfoProvider requires a string here as input (but does not use it)

src/passport-saml/saml.ts
gugu: returning value depends on input string, typescript does not know about `/@InRenponseTo` response. Updated to `as Attr[]`

src/passport-saml/saml.ts
gugu: express has the support of parsing multiple Signature parameters. Do we need to write code to cover every case like this? It will be tens of Array.isArray() conditions.

markstos:
As a practical matter, we got multiple messages over the past 12 hours about bugs in the current release and people asking for a new fixed release, so I'm inclined to merge this in the next few hours and continue to improve our TypeScript support in future releases.

src/passport-saml/saml.ts
```diff
@@ -8,18 +8,27 @@ import * as xmldom from 'xmldom';
...
 import * as xmlbuilder from 'xmlbuilder';
+// @ts-expect-error no typings for xml-encryption
 import * as xmlenc from 'xml-encryption';
 import * as util from 'util';
 import {CacheProvider as InMemoryCacheProvider} from './inmemory-cache-provider';
 import * as algorithms from './algorithms';
 import { signAuthnRequestPost } from './saml-post-signing';
 import type { Request } from 'express';
+import { ParsedQs } from 'qs';
+import { AudienceRestrictionXML, AuthenticateOptions, AuthorizeOptions, AuthorizeRequestXML, CertCallback, LogoutRequestXML, Profile, RequestWithUser, SamlScopingConfig, ServiceMetadataXML, XMLInput, XMLObject, XMLOutput, XMLValue } from './types';
```
markstos, line 19: Is it a mistake that that `import` is called on the `types` file twice?

I'm working on merging this branch and plan to simplify this.

markstos:
This has been rebased and merged through another branch, so closing this.

cjbarth:
> Let's keep a balance between strict types and readability of the code, that's the awesome part of TypeScript. Sometimes it is better to add `!` or force type casting instead of writing code just to make typescript happy

It really isn't better to do that... really, it isn't. The much better alternative is to leave off types entirely and not be so strict with the type system. Leaving off types tells the developer that it could be anything; putting on the wrong type tells the developer that you've done your homework and found it be be such-and-such... and in fact, you're lying. It is so much harder to debug code with the wrong types than with no types at all. It is bad to ruin readability of code just to make TypeScript happy, but it is also bad to add lies about the code types to make TypeScript happy. This is the entire reason that there are so many strictness options in TypeScript type checking.

Having said that, I'll be glad to help review code to be more realistic with our types. You've done a lot of good work here and I want to try to keep up with you and provide solid experienced feedback (been in TypeScript since 1.0 some 6 years ago). Strict for the sake of strict is not a good policy when it results in adding inaccurate types like `!` just to be able to turn on a type checking feature; it is lying on a mortgage application to get a loan.

src/passport-saml/saml.ts
cjbarth: No, that would be the wrong approach. Really, what we want is to not assume that we have a string, but tell the user when we don't have a string, so all we need is this:

```javascript
if (typeof container.Signature === 'string' && typeof container.SigAlg === 'string') {
    return this.validateSignatureForRedirect(
        urlString, container.Signature, container.SigAlg, cert
    );
} else {
    throw new Error("Unexpected Signature or Signature Algorithm format.")
}
```

That really increases the quality of the code and helps the user. Simply adding `as string` hurts the developer and doesn't help the user.

markstos:
@cjbarth Thanks for volunteering to help refine. I'm aligned with your perspective.

PR #487, opened 2020-11-02
inlineSources option for better source maps

gugu:
Typescript source code is inlined inside source map


markstos:
@gugu This seems fine. I read it requires `sourceMap` or `inlineSourceMap` to be set and I confirmed that `sourceMap` is set.

cjbarth:
I'm not accustomed to having source maps inline for production systems for lots of little (probably mostly trivial reasons). Is there a reason we prefer inline source maps?

PR #488, opened 2020-11-02
Allow for use of privateKey instead of privateCert

alon85:
An updated PR to address the conversation here: https://github.com/node-saml/passport-saml/pull/387

### DIFF
--- a/src/passport-saml/saml-post-signing.ts
+++ b/src/passport-saml/saml-post-signing.ts
@@ -6,7 +6,8 @@ const issuerXPath = '/*[local-name(.)="Issuer" and namespace-uri(.)="urn:oasis:n
 const defaultTransforms = [ 'http://www.w3.org/2000/09/xmldsig#enveloped-signature', 'http://www.w3.org/2001/10/xml-exc-c14n#' ];
 
 interface SignSamlPostOptions {
-  privateCert: string;
+  privateCert?: string;
+  privateKey?: string;
   signatureAlgorithm?: string;
   xmlSignatureTransforms?: string[];
   digestAlgorithm: string;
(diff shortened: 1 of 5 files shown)


markstos (approved):
This has code/docs/tests and is backward compatible. It looks good to me. I updated the docs to further simplify them and focus on the clearer `privateKey` name instead of the less clear "private cert". 

Assigning to @cjbarth for a second look.

src/passport-saml/saml-post-signing.ts
```diff
@@ -15,15 +16,15 @@ interface SignSamlPostOptions {
 export function signSamlPost(samlMessage: string, xpath: string, options: SignSamlPostOptions) {
   if (!samlMessage) throw new Error('samlMessage is required');
   if (!xpath) throw new Error('xpath is required');
-  if (!options || !options.privateCert) throw new Error('options.privateCert is required');
+  if (!options || (!options.privateCert && !options.privateKey)) throw new Error('options.privateCert or options.privateKey is required');
```
cjbarth, line 19: What do you think of issuing a deprecation warning if someone is using `privateCert`? This way, we can remove this extra code with a future semver-major release.

cjbarth (approved):
I just have one suggestion, but I don't think it should block this merge.

RyanCopley (approved):
:wave: Thank you Alon! I hope you are doing well :)

src/passport-saml/saml-post-signing.ts
markstos: @cjbarth That's OK if you want to add the warning.

markstos:
I'm going to go ahead and merge this. We can add a deprecation warning in a future release. It will clean up the code to eventually not support both names.

PR #476, opened 2020-10-29
compatibility with @types/passport-saml, fixes #475

gugu:
# Description
Before I did not include typings in package.json, but typescript still detects new definitions and turns off @types/passport-saml package

This PR adds compatibility with @types/passport-saml

 - Issue Addressed: [#475]
 - Tests included? [in #475]

Issue #489, opened 2020-11-03
New release?


robcresswell:
Could we get a new release tagged with the latest fixes please?

markstos:
In progress. I'll leave a comment here when it's out.

markstos:
Published v2.0.1. 

Release notes are here: https://github.com/node-saml/passport-saml/releases/tag/v2.0.0

robcresswell:
Thanks @markstos! Much appreciated

PR #490, opened 2020-11-03
fix(typing): multi saml strategy typings

rob-gijsens:
Hi,

I made an fix to the typings for the MultiSamlStrategy based on the changes recently merged with PR #476 .
Without this fix the MultiSamlStrategy can't be imported in my Typescript project without an error.

Besides the fix as stated above I noticed the semver number isn't corresponding with the latest release (1.4.2 in the package.json file in contrast to the latest release 1.5.0). I updated the package.json to 1.5.0 so it contains at least the same version number as the latest release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "passport-saml",
-  "version": "1.4.2",
+  "version": "1.5.0",
   "license": "MIT",
   "keywords": [
     "saml",
--- a/src/passport-saml/index.ts
+++ b/src/passport-saml/index.ts
@@ -1,6 +1,7 @@
 import type { CacheItem, CacheProvider} from './inmemory-cache-provider';
 import { SAML } from './saml';
 import Strategy = require('./strategy');
+import MultiSamlStrategy = require('./multiSamlStrategy');
 import type { Profile, VerifiedCallback, VerifyWithRequest, VerifyWithoutRequest } from './types';
 
-export { SAML, Strategy, CacheItem, CacheProvider, Profile, VerifiedCallback, VerifyWithRequest, VerifyWithoutRequest };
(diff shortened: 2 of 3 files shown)


gugu:
fixed in #485 with multiSamlStrategy typing

rob-gijsens:
@gugu Thanks for your reply. 
I noticed your PR #485 and the fact that this fixes the same issue. But as you stated in your PR, your change is probably a major version change, In my opinion it is nice to also add my (simpler) fix to the v1.x branch for anyone who doesn't want to upgrade a major version immediately, but do want to use the multi saml strategy. What are your thoughts on this?

markstos:
@rob-gijsens The 2.x series should not require major code changes, and we have limited volunteer maintainers. You are welcome to fork the 1.x series and maintain a patched version of that, but I'm not inclined to support a second branch in this project.

Changes required to update to 2.x will be updated in the Release notes and will be minor for most people and no changes at all for some people.

rob-gijsens:
@markstos Fair enough.
Is there at least an indication for when the 2.x version is released?
I recently updated passport-saml in my project from 1.3.3 to 1.5.0 due to a  [vulnerability in the 1.3.3 release caused by the dependency on xml-crypto.](https://snyk.io/test/npm/passport-saml/1.3.3). But since the update to a more recent version of passport-saml where the vulnerable version of xml-crypto is updated, my multi SAML strategy isn't working anymore due to the bug I fixed in this PR.

I'm kinda stuck in the middle right now...

markstos:
Today. Subcribe to #489 for notification.  

I think you are probably not vulnerable to the `xml-crypto` bug if you are not using HMAC-SHA1, which `passport-saml` does not use by default.

PR #492, opened 2020-11-03
Add deprecation notice for privateCert; fix bug

cjbarth:
This adds the deprecation notice for `privateCert` in favor of `privateKey`. It also fixes a bug introduced in the refactoring of `privateCert` to `privateKey`.


markstos (approved):
Thank you, this is cleaner.

PR #493, opened 2020-11-03
Prettier + ESLint + onchange = Happiness

cjbarth:
This adds support for `prettier` working along side `eslint`. It doesn't enable anything.

I'm marking this as draft because there is a problem with some variable replacement in one watch command that I'm working on, but I wanted to get this out there for review. I imagine that there will be a follow-on PR after this lands that will reformat all the code in one commit.


.eslintrc
```diff
@@ -0,0 +1,18 @@
...
+    "eslint:recommended",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier"
+  ],
+  "rules": {
+    "no-console": 1,       // Means warning
+    "prettier/prettier": 1 // Warning for now, until we get this cleaned up
+  }
+}
```
gugu, line 18: I think we need to remove eslint config from package.json in this case

cjbarth: Good point 👍 

May I ask what the rationale behind these rules was? I'm wondering if I should move them over or not.

```
    "no-undefined": "off",
    "no-unused-vars": "off",
```

markstos:
Good progress. We'll save this for a future 2.x release

.eslintrc
gugu: image

I'm not the right person for this question) I think these rules should not be disabled, they are helpful

cjbarth:
This PR is ready to be reviewed/merged now.

gugu (approved):

PR #495, opened 2020-11-03
provide path to js in main entry, fixes #494

### LINKED ISSUE #494: [BUG] Cannot find module '/www/node_modules/passport-saml/lib/passport-saml'. Please verify that the package.json has a valid "main" entry
Still getting the same error in 2.0.1 that I was in 1.5.0. Anyone else having this issue?

```internal/modules/cjs/loader.js:329
      throw err;
      ^

Error: Cannot find module '/www/node_modules/passport-saml/lib/passport-saml'. Please verify that the package.json has a valid "main" entry
    at tryPackage (internal/modules/cjs/loader.js:321:19)
    at Function.Module._findPath (internal/modules/cjs/loader.js:534:18)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:27)
    at Function.Module._load (internal/modules/cjs/loader.js:743:27)
    at Module.require (internal/modules/cjs/loader.js:965:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/www/dist/backend/index.js:36:36)
    at Module._compile (internal/modules/cjs/loader.js:1076:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Module.load (internal/modules/cjs/loader.js:941:32)
    at Function.Module._load (internal/modules/cjs/loader.js:782:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  path: '/www/node_modules/passport-saml/package.json',
  requestPath: 'passport-saml'
}
```

gugu:


package.json
```diff
@@ -34,7 +34,7 @@
     "docs",
     "package-lock.json"
   ],
-  "main": "./lib/passport-saml",
+  "main": "./lib/passport-saml/index.js",
```
cjbarth, line 37: I'm OK with this change, and I actually expected this to be this way from the beginning. I guess I'm curious why everything has been working for so long with an abstract `main` (I checked the history) and now it is broken.

cjbarth (approved):

markstos:
@gugu did you test the fix with Babel? I don't think is this the issue. If you look at the JavaScript we generate in `../lib/passport-saml/index.js` you can see it contains `__esModule: true`. If you look the code in #494, you can see references to `cjs`, or Common JS modules. So it seems like we have perhaps there's a mismatch between an expectation about whether the module is a Common JS module or an ES 6  module. There may be specific issues related to this topic and Babel:

https://2ality.com/2017/01/babel-esm-spec-mode.html

It seems Babel may be looking for `__esModule` in particular (it's mentioned at the URL above).

I'm not an expert in TypeScript, Babel and how these interact with Common JS and ES 6 modules, but that seems to the be area where the issue lies.

CC: @alon85

markstos:
The actual fix as posted here: https://github.com/node-saml/passport-saml/issues/494#issuecomment-721699139

PR #504, opened 2020-11-19
docs(scoping): fix for example

rob-gijsens:
Fix in the README example of the scoping option which I previously added in a different PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,16 +137,18 @@ type Profile = {
   * `scoping`: An optional configuration which implements the functionality [explained in the SAML spec paragraph "3.4.1.2 Element <Scoping>"](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf). The config object is structured as following:
 ```javascript
 {
-  idpList: { // optional
-    entries: [ // required
-      {
-        providerId: 'yourProviderId', // required for each entry
-        name: 'yourName', // optional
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -137,16 +137,18 @@ type Profile = {
...
 ```javascript
 {
-  idpList: { // optional
-    entries: [ // required
-      {
-        providerId: 'yourProviderId', // required for each entry
-        name: 'yourName', // optional
-        loc: 'yourLoc', // optional
-      }
+  idpList: [ // required
```
cjbarth, line 140: Is `idpList` really required now? If so, I think that may have been a breaking change that we didn't properly account for.

rob-gijsens: @cjbarth My mistake.

The idpList isn't required as stated in the saml2.0 spec. It actually wasn't required in the previously added feature either as can be seen in `/src/passport-saml/saml/ts:361`.
The only mismatch was in the Typescript interface `SamlScopingConfig`. I reverted the required to optional in the documentation and fixed the Typescript interface in my latest commit.

cjbarth (approved):

rob-gijsens:
@cjbarth Thanks for approving. What is next in the PR process? I don't have write access to this repository to merge this PR myself, and also haven't the right authorization to assign someone who does. Is this automatically being merged when a next release is made?

PR #507, opened 2020-12-03
Fix types for inmemory cache provider

gugu:
Types are fixed for cache provider. It currently works okay, but if somebody tries to override, it display incorrect types.

BTW: let's consider displaying some warning in README about inmemory cache provider, which is kind of unusable in production


gugu:
I think it is better to make an interface and don't touch in memory cache provider

PR #508, opened 2020-12-04
add ts-ignore to generated type definitions for multisaml strategy

gugu:
# Description

1) Currently, we are forced to use ts-expect-error for multi-saml configuration because it extends the method in the base class with a completely different signature. But this ts-expect-error was not propagated to typescript definitions files. This PR makes things even worse - it uses undocumented feature of TS to propagate ts-expect-error in the definitions
2) to be able to use passport-saml from github I've changed tsc command from `prepublishonly` to `prepack` step. In this case tsc will run when package is installed directly from github

In v3 I suggest to change `generateServiceProviderMetadata` in multisaml strategy to `generateServiceProviderMetadataAsync` and add a stub with error message fro `generateServiceProviderMetadata`

# Checklist:

 - Issue Addressed: [ ]
 - Link to SAML spec: [ ]
 - Tests included? [ ]
 - Documentation updated? [ ]

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "lint": "eslint --ext .ts src",
     "lint-watch": "onchange 'src/**/*.ts' -- eslint {{file}}",
     "lint:fix": "eslint --ext .ts --fix src",
-    "prepublishOnly": "tsc",
+    "prepare": "tsc",
     "prettier-check": "prettier --config .prettierrc --check .",
     "prettier-format": "prettier --config .prettierrc --write .",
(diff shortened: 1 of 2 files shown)


cjbarth (approved):
I really would like to see us getting rid of all the "expected" TS errors. TS has grown to the point where manually saying what types to expect or what errors to expect is probably something that we can avoid.

cjbarth:
I'm not thrilled about the "v3" idea you've presented, but that can be a discussion for another time.

gugu:
@cjbarth yes, I don't like v3 idea as well and I wish there is any way to solve the problem better way

PR #510, opened 2020-12-07
types: avoid ts-expect-error on generateServiceProviderMetadata

midgleyc:
# Description

In multiSamlStrategy, generateServiceProviderMetadata currently changes the signature from the base class, and uses a ts-expect-error so that compilation continues to work. However, the type files show errors because the subclass is no longer an instance of its base class.

This adds an overload to generateServiceProviderMetadata taking the original arguments and returning "never", while the implementation still checks the type of the callback to confirm we're using the right call.

# Checklist:

 - Issue Addressed: [x]
 - Link to SAML spec: [ ]
 - Tests included? [ ]
 - Documentation updated? [ ]

### DIFF
--- a/src/passport-saml/multiSamlStrategy.ts
+++ b/src/passport-saml/multiSamlStrategy.ts
@@ -57,18 +57,24 @@ class MultiSamlStrategy extends SamlStrategy {
     });
   }
 
-  /** @ts-expect-error typescript disallows changing method signature in a subclass */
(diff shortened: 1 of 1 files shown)


gugu:
I vote for #508 as it does not alters the base class

gugu:
#508 is merged in master branch

cjbarth:
@midgleyc , did #508 fix your problem? Can this be closed?

midgleyc:
Sure, ignoring the error in the type file too works.

Issue #511, opened 2020-12-11
[BUG] Signature calculation is inconsistent for different line endings


mhassan1:
In https://github.com/node-saml/passport-saml/pull/498, we added newline normalization before signature validation in `validateSignatureForCert`. The suggestion for this fix came from https://github.com/node-saml/passport-saml/issues/431#issuecomment-629170914; however, I don't believe this fix goes deep enough into newline normalization.

While this fix normalizes newlines right before the signature validation, it doesn't normalize newlines in the XML payload that is passed to `sig.loadSignature`. As a result, the `sig.getCanonSignedInfoXml` that is part of `sig.checkSignature` encodes carriage returns into `&#xD;`, and the signature computed from that canonicalized SignedInfo XML is incorrect.

** Spec-driven development **

This is related to the XML spec for newlines mentioned here: https://github.com/node-saml/passport-saml/issues/431#issuecomment-718132752

** Community development model **

PR to come

**To Reproduce**

1. Add a newline to `response.root-signed.assertion-signed.xml` inside the `Signature` block.
2. Re-run `test-signatures.js` and see that `R1A - both signed => valid` fails, as expected, since the signature should be different now that the XML is different.
3. Compute the correct digests and signatures for this modified file and update them in `DigestValue` and `SignatureValue`
4. Re-run `test-signatures.js` and see that all tests pass except for `CRLF line endings` and `CR line endings`.

**Expected behavior**
XML payloads with non-LF line endings should have the same digest and signature values as the same XML payload with LF line endings, and it is up to the XML processor to normalize line endings to ensure that (https://github.com/node-saml/passport-saml/issues/431#issuecomment-718132752)

**Environment**
 - Node.js version: 14.15.1
 - passport-saml version: 2.0.2

PR #512, opened 2020-12-11
normalize signature line endings before loading signature block to xml-crypto

mhassan1:
# Description

Resolves https://github.com/node-saml/passport-saml/issues/511

# Checklist:

 - Issue Addressed: [X]
 - Link to SAML spec: [ ]
 - Tests included? [X]
 - Documentation updated? [ ]

### DIFF
--- a/src/passport-saml/saml.ts
+++ b/src/passport-saml/saml.ts
@@ -733,6 +733,7 @@ class SAML {
       getKeyInfo: key => "<X509Data></X509Data>",
       getKey: keyInfo => Buffer.from(this.certToPEM(cert)),
     };
+    signature = this.normalizeNewlines(signature.toString());
     sig.loadSignature(signature);
     // We expect each signature to contain exactly one reference to the top level of the xml we
     //   are validating, so if we see anything else, reject.
@@ -752,7 +753,7 @@ class SAML {
     if (totalReferencedNodes.length > 1) {
       return false;
     }
-    fullXml = fullXml.replace(/\r\n?/g, '\n');
+    fullXml = this.normalizeNewlines(fullXml);
     return sig.checkSignature(fullXml);
   }
 
@@ -1432,6 +1433,13 @@ class SAML {
       ''
     ].join('\n');
     return wrappedKey;
+  }
+
+  normalizeNewlines(xml : string) : string {
+    // we can use this utility before passing XML to `xml-crypto`
+    // we are considered the XML processor and are responsible for newline normalization
+    // https://github.com/node-saml/passport-saml/issues/431#issuecomment-718132752
+    return xml.replace(/\r\n?/g, "\n");
   }
 }
 
--- a/test/static/signatures/valid/response.root-signed.assertion-signed.xml
+++ b/test/static/signatures/valid/response.root-signed.assertion-signed.xml
@@ -35,5 +35,35 @@
                 </saml:AttributeValue>
             </saml:Attribute>
         </saml:AttributeStatement>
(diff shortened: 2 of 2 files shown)


test/test-signatures.js
```diff
@@ -41,7 +41,7 @@ describe('Signatures', function() {
 
   describe('Signatures on saml:Response - Only 1 saml:Assertion', () => {
     //== VALID
-    it('R1A - both signed => valid', testOneResponse('/valid/response.root-signed.assertion-signed.xml', false, 1));
+    it('R1A - both signed => valid', testOneResponse('/valid/response.root-signed.assertion-signed.xml', false, 2));
     it('R1A - root signed => valid', testOneResponse('/valid/response.root-signed.assertion-unsigned.xml', false, 1));
```
mhassan1, line 45: I'm not sure why adding newlines to the SAMLResponse causes this change to be needed, but it does

test/static/signatures/valid/response.root-signed.assertion-signed.xml
```diff
@@ -35,5 +35,35 @@
...
+            <ds:Reference URI="#_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa">
+                <ds:Transforms>
+                    <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
+                    <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
+                </ds:Transforms>
+                <ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"/>
+                <ds:DigestValue>wL0JrPFcOmwrYpIQuLBuT3kYdFr4kXy5yrrhlxpZK6k=</ds:DigestValue>
+            </ds:Reference>
+        </ds:SignedInfo>
+        <ds:SignatureValue>XcZR4j7Hy7QQPoYuAlIi4k/6BK+Nq/ByyPOzIdlkMD4BT7HobxgMv0LeFvgdvE0QMA1IUQgGVuMKRMdwy02HowhQHPrLzHomwKXIy0GaBuDKudLcjYH4Sqlo1DMn5MHhXyzE4j66FlJFoR0fvjWyz4vbRHFpxMcVoqrb4NBDT5f5LN63IXlexZOOCR0n6ZimyNpu91LzCtK4mYb53OjT1Glbq3zslxas+jqEk4+S2wqZyIQjqzsUrOl540L9+9x/KlfKB2jeb2i2wjiwLORzzlZXPrOxLZzisZMnossS9QXImj9ItF9vE9coIJGbF30YSSmGeLgdxjRLJGQFdg5dPg==</ds:SignatureValue>
```
mhassan1, line 67: I'm not sure why this signature value stayed the same with the addition of these newlines, while the assertion signature value changed

test/test-signatures.js
cjbarth: Since this is changing a test, which is basically our contract to our consumers that we haven't broken anything or change any behavior unexpectedly, we really should know why this value much change, or we'll have to do a breaking change release. Even then, we should know _why_ the change is needed.

cjbarth (changes requested):

test/static/signatures/valid/response.root-signed.assertion-signed.xml
mhassan1: Nevermind, this signature value has now been updated correctly

test/test-signatures.js
mhassan1: Nevermind, I see why. I did not update the signature value in the fixture correctly.

src/passport-saml/saml.ts
```diff
@@ -733,6 +733,7 @@ class SAML {
       getKeyInfo: key => "<X509Data></X509Data>",
       getKey: keyInfo => Buffer.from(this.certToPEM(cert)),
     };
+    signature = this.normalizeNewlines(signature.toString());
```
mhassan1, line 736: You can verify this fix by removing this line and running `test-signatures.js`. The `CRLF line endings` and `CR line endings` tests will fail, while the `R1A - both signed => valid` will pass, even though all three tests use the same fixture. This normalization of newlines passes all three tests.

cjbarth (approved):

mhassan1:
@cjbarth Is there anything I can do to help get this merged? What is your expectation on a release timeline?

cjbarth:
@mhassan1 We don't have a release timeline. It is nice when people can use the specific commit in their `package.json` to test the latest code. After a period of stability, we'll typically cut a release. @mans0954 has been doing a lot of the releases, and I like lots of eyes on security stuff, so Iet's wait for his thoughts on this matter.

markstos:
This looks good to me. @cjbarth I'm fairly fair behind on my passport-saml mail at this point. 

If you want to go ahead cut the next the release that's fine with me. I believe you have NPM publish rights.

mhassan1:
@cjbarth Are you planning a new release? One problem with pointing to a specific commit is that the compiled JS is not checked in, so we would need to build the project ourselves.

cjbarth:
Some recent changes that we made should cause a compilation when it is installed using a specific commit. Is that not working?

mhassan1:
How should that work? Which recent change? We are using `yarn v2`, which will run a `postinstall` step, if it exists, but this package doesn't have one (and that's discouraged just for the purpose of compilation).

mhassan1:
Oh, I see the prepare script. I would still rather point to an NPM dist if you are planning to cut a release in the next few days. Are you planning to? If not, please let me know so I can plan accordingly.

PR #515, opened 2020-12-15
fix: derive SamlConfig from SAMLOptions

midgleyc:
Strategy and MultiSamlStrategy accept an "options" parameter with three fields which are used locally by the strategy, and remaining fields which are passed on to the SAML constructor.

There was an issue where certain fields (e.g. `privateKey`) could not be passed to the Strategy constructor but had to be passed on to the SAML constructor. This fixes that more thoroughly than #497, also fixing the type of `authnContext`, the limited string parameters, and adding `xmlSignatureTransforms`, `digestAlgorithm` and `disableRequestACSUrl`. 

# Checklist:

 - Issue Addressed: [x]
 - Link to SAML spec: [ ]
 - Tests included? [ ]
 - Documentation updated? [ ]

### DIFF
--- a/src/passport-saml/saml-post-signing.ts
+++ b/src/passport-saml/saml-post-signing.ts
@@ -1,6 +1,6 @@
 import { SignedXml } from 'xml-crypto';
 import * as algorithms from './algorithms';
-import { SAMLOptions } from './saml';
+import { SAMLOptions } from './types';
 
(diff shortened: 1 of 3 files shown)


cjbarth (approved):

cjbarth:
@gugu Since you were involved in a lot of the TypeScript conversions, I'd like you to have a look at this too. It seems good to me, but 4 eyes can't hurt :)

gugu:
Looks good

PR #518, opened 2021-01-07
Generating changelog using gren

gugu:
We can get it here: https://github.com/github-tools/github-release-notes


CHANGELOG.md
```diff
@@ -0,0 +1,581 @@
+# Changelog
+
+## v2.0.3 (01/01/1970)
```
cjbarth, line 3: This date is wrong

CHANGELOG.md
```diff
@@ -0,0 +1,581 @@
...
+- [**closed**] normalize signature line endings before loading signature block to xml-crypto [#512](https://github.com/node-saml/passport-saml/pull/512)
```
cjbarth, line 4: Do we have to have the status here? Some of these status and tags really add nothing to the change log IMO.

cjbarth: Maybe we can use this config value to remove the tags: https://github-tools.github.io/github-release-notes/options.html#ignore-labels

.grenrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
cjbarth, line 1: It seems we might benefit from quickly tagging recent commits so we can use grouping. Thoughts? https://github-tools.github.io/github-release-notes/options.html#group-by-1

cjbarth (changes requested):
Looks really nice. This should speed the process going forward. I see we can even put a badge if we wanted! https://github.com/github-tools/github-release-notes#setup

gugu:
@cjbarth can you please look changelog one more time? Currently, it includes correct changes

CHANGELOG.md
cjbarth: Can the date format be ISO 8601 standard? E.g. 2021-01-07

CHANGELOG.md
```diff
@@ -0,0 +1,525 @@
...
+- [**closed**] Fix broken tests [#367](https://github.com/node-saml/passport-saml/pull/367)
+- [**peer-review-welcome**] Create a way to get provider metadata when using the MultiSamlStrategy [#323](https://github.com/node-saml/passport-saml/pull/323)
+- [**closed**] Extend and document the profile object [#301](https://github.com/node-saml/passport-saml/pull/301)
+- [**pending-refinement**] feat: add RequestedAuthnContext Comparison Type parameter [#360](https://github.com/node-saml/passport-saml/pull/360)
+- [**closed**] Update README.md [#363](https://github.com/node-saml/passport-saml/pull/363)
+- [**peer-review-welcome**] InResponseTo support for logout [#356](https://github.com/node-saml/passport-saml/pull/356)
+
+---
+
+## v0.32.0 (01/01/1970)
```
cjbarth, line 385: This seems both out-of-place and wrong.

CHANGELOG.md
gugu: it will require custom template with date parsing/formatting code, there is no option for date format

cjbarth: It seems that way, sad... thankfully, someone already wrote the template, and all we have to do is paste it in :)

https://github.com/github-tools/github-release-notes/issues/147#issuecomment-584759005

.grenrc.js
```diff
@@ -0,0 +1,8 @@
...
+    "dataSource": "prs",
+    "prefix": "",
+    "onlyMilestones": false,
+    "ignoreTagsWith": ["v0.32.0", "0.0.3"],
```
cjbarth, line 5: Why are these ignored? Is that just for immediate expediency? We can always add a tag and then tell it to only work with tags with a "v" prefix if that helps.

gugu: These tags were created but no release was published. Changelog for these tags will be broken

gugu: and there is no chance to fix it because creating releases for old tags is not possible

CHANGELOG.md
gugu: image

.grenrc.js
gugu: I tried, but usually we have up to 10 PRs per release, grouping creates a group per 1-2 PRs

cjbarth: We can always revisit this later too.

cjbarth:
Final question: is the script in `package.json` the one that generates this `CHANGELOG.md`? If not, could you please put the script you use there so that others can duplicate your work later on?

cjbarth:
I'm not sure what I'm doing wrong then, because even that change you've made isn't working for me. I'm getting the correct tags, but the date is always 1969-12-31 and the message is always "*No changelog for this release.*". Given that we already have `tags: "all"` in the config file and `-G` is the same as `--generate`, I feel like something else is amiss here. You have this all working properly with `npm run changelog`?

cjbarth (approved):

cjbarth:
I got it working reliably. I also use `prettier` on it, as I'd like to roll that out soon as we decide on the async/await thing; I don't want to break the merge too badly by using `prettier` before then. Feel free to merge this if you are happy with what I've done. Thanks for your hard work on this.

Issue #525, opened 2021-01-29
[BUG]


dosullivan557:
When installing the module, we pull in .pem files and .key files which causes our application to fail security scans. As part of the install, we should not move these files in.

cjbarth:
I'm not sure what bug you are reporting. Can you please provide a clear test case, perhaps even a new unit test that fails?

dosullivan557:
It's not an issue with the code, because we install your module in an application, it copies over .pem and .key files that are in your tests. This means that when we put our application through security scans, it fails because it finds these files and classifies them as sensitive data. What I'm asking is whether you can not include these in the bundled version of the application.

To resolve it, you can have a file called .npmignore and specify the paths, which will then not copy those files over to the build of the module.

cjbarth:
That seems reasonable. Can you make a PR?

dosullivan557:
Thanks @cjbarth  - PR raised

dosullivan557:
Merged in - @cjbarth when could I expect it to be release to use?

cjbarth:
Released as v2.0.5

PR #526, opened 2021-01-29
Ignore `test` folder when building npm package

dosullivan557:
# Description

Thank you for taking the time to create a PR to support this community-maintained project. We are all volunteers here, so features are only added by contributers like you. Please also be patient while the maintainers work to review your proposed changes for quality and effacy.

Please include a _summary of the change_ and, where appropriate, _which issue is addressed_.

Your description should also include a _comment about the use-case_ for this change, including what systems you may be interfacing with that require this change. This will help future users know how this library is being used in the wild and how well supported it is.

If the change involves the SAML specification, please _include a link to the relavent part(s) of the SAML specificaiton_. Doing so speeds up the review process and helps the maintainers make sure that this project remains spec-compliant. You might start [here](https://www.oasis-open.org/standards#samlv2.0) to find the part of the specificaiton that relates to this PR.

_Please include tests._ Doing so will ensure that the changes made in this PR are not undone or otherwise corrupted by future changes.

# Checklist:

 - Issue Addressed: [ ]
 - Link to SAML spec: [ ]
 - Tests included? [ ]
 - Documentation updated? [ ]


dosullivan557:
resolves #525

.npmignore
```diff
@@ -0,0 +1,3 @@
+**/**/.certs
```
cjbarth, line 1: Why not just ignore the entire tests folder? That doesn't really need to be in the `npm` package.

cjbarth (changes requested):

.npmignore
dosullivan557: Updated the PR to not copy the test folder over

cjbarth (approved):

PR #530, opened 2021-02-02
Fix typing export for MultiSAMLStrategy

thanhvg:
otherwise `tsc` will complain about nonsense import when used in a project

``` 
node_modules/passport-saml/multiSamlStrategy.d.ts:1:36 - error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.

1 import * as MultiSAMLStrategy from './lib/passport-saml/multiSamlStrategy';
```

### DIFF
--- a/multiSamlStrategy.d.ts
+++ b/multiSamlStrategy.d.ts
@@ -1,2 +1,2 @@
-import * as MultiSAMLStrategy from "./lib/passport-saml/multiSamlStrategy";
+import MultiSAMLStrategy = require("./lib/passport-saml/multiSamlStrategy");
 export = MultiSAMLStrategy;


gugu:
This code is a copy of what we had before. Changing this line can break the code for JS users

 I suggest to update export in src/multisamlstrategy.ts, keep exports in this file as is and update imports in user's code to using lib/multisamlstrategy.ts

cjbarth:
@gugu , do you think the problems we had before could be partially due to using `export =` instead of a named export?

gugu:
@thanhvg can you please update import in your code to 
`import {MultiSamlStrategy} from 'passport-saml'` ?

cjbarth:
Thanks for following on this @gugu . It appears that your work on #531 will help resolve this. I look forward to feedback from @thanhvg .

thanhvg:
> `import {MultiSamlStrategy} from 'passport-saml'` ?

This one works, thank you all for your help. I'm closing this PR.

PR #531, opened 2021-02-04
Update readme on using multiSamlStrategy

gugu:
passport-saml/multiSamlStrategy.js is something I'd like to remove:
1. it contains nothing, just import
2. it is outside of the source root and tsc/other tools don't include this file
3. old import depends on internal structure of the project
4. it uses module-level export

For now let's keep the file for compatibility, but let's update README so new people will not use the old way

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The examples utilize the [Feide OpenIdp identity provider](https://openidp.feide
 The SAML identity provider will redirect you to the URL provided by the `path` configuration.
 
 ```javascript
-var SamlStrategy = require('passport-saml').Strategy;
+const SamlStrategy = require('passport-saml').Strategy;
 [...]
 
 passport.use(new SamlStrategy(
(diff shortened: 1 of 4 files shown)


cjbarth (approved):

cjbarth:
I feel like we can remove that old file for a v3 release, which I believe is the next in line, so I say, remove it.

gugu:
@cjbarth done

markstos:
We just are getting to upgrading to 3.x ourselves and although I read through the CHANGELOG for breaking changes, I missed this. It could use a more clear note in the Changelog that it's a breaking change and a code change is required:

```diff
-var MultiSamlStrategy = require('passport-saml/multiSamlStrategy');
+const { MultiSamlStrategy } = require('passport-saml');
```

However, given the automated Changelog generation I'm not sure how to add that. At least if someone follows a link from the Changelog to here, they'll find the required syntax change more readily.

cjbarth:
The PR that caused the break can be tagged with `breaking-change` and then it will get re-organized in the automatic changelog generation.

PR #532, opened 2021-02-09
async / await in cache interface

gugu:
Currently, the cache interface is a very important part of this module because the default one is not usable in the production environment.

For compatibility reasons the interface was kept callback-based, despite the module itself uses async/await under the hood. For v3 I propose this breaking change to simplify our and users' code

### DIFF
--- a/src/passport-saml/inmemory-cache-provider.ts
+++ b/src/passport-saml/inmemory-cache-provider.ts
@@ -48,7 +48,7 @@ export class CacheProvider {
           nowMs >=
           new Date(this.cacheKeys[key].createdAt).getTime() + this.options.keyExpirationPeriodMs
         ) {
-          this.remove(key, () => undefined);
+          this.removeAsync(key);
         }
       });
     }, this.options.keyExpirationPeriodMs);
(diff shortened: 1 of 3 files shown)


src/passport-saml/inmemory-cache-provider.ts
```diff
@@ -48,15 +48,15 @@ export class CacheProvider {
           nowMs >=
           new Date(this.cacheKeys[key].createdAt).getTime() + this.options.keyExpirationPeriodMs
         ) {
-          this.remove(key, () => undefined);
+          this.removeAsync(key);
```
cjbarth, line 51: I understand that it is customary to use the `Async` suffix for Promise-returning functions, however, that seems to be an unnecessary breaking change as it would completely change the API for cache providers, which could break a lot of plugged-in code. Could we not rename the function, or would it be advisable to not append `Async` since we aren't actually making this newly async as it always was that way, just with callbacks?

cjbarth (changes requested):
Some initial comments after a brief review. I'll look more closely later.

src/passport-saml/inmemory-cache-provider.ts
gugu: If I don't change method names, the problem will be much larger - cache functions will be called, but callback - will not. Instead of crashing on compile-time people will get a timeout in their production code

cjbarth (approved):

cjbarth:
Please use a "Squash and merge" when merging to keep the git history cleaner.

PR #534, opened 2021-02-10
Tests use typescript

gugu:
Tests switched to ts.

1. PR includes async cache changes for now
2. I did not add types for the request module - we need to remove this dinosaur instead of adding types to it


cjbarth:
I'll review this when the cache code is merged and this is rebased on top of it.

gugu:
@cjbarth done

test/test-signatures.ts
```diff
@@ -27,21 +31,23 @@ describe("Signatures", function () {
         //== Run the test in `func`
         samlObj.validatePostResponse(
           samlResponseBody,
-          tryCatchTest(done, function (error) {
+          tryCatchTest(done, function (error: any) {
             //== Assert error. If the error is `SAML assertion expired` we made it past the certificate validation
-            shouldErrorWith
-              ? error.should.eql(new Error(shouldErrorWith))
-              : error.should.eql(new Error("SAML assertion expired"));
+            error.should.eql(new Error(shouldErrorWith || "SAML assertion expired"));
```
cjbarth, line 36: The way that this has been refactored, while it still works (in a non-obvious way), now seems to make the comment less useful. It seems that we should know when to expect which item in the test instead of saying both options are expected here.

test/tests.ts
```diff
@@ -1111,70 +1157,76 @@ describe("passport-saml /", function () {
...
-        var userSerialized = false;
+        let userSerialized = false;
         passport.serializeUser(function (user, done) {
           userSerialized = true;
           done(null, user);
         });
 
         app.get(
           "/login",
+          // @ts-expect-error the problem of passport typings
```
cjbarth, line 1186: Is this something I can help fix over at DefinitelyTyped?

test/tests.ts
```diff
@@ -2167,86 +2229,90 @@ describe("passport-saml /", function () {
...
-        var samlConfig = {
+        const samlConfig = {
           entryPoint: "",
           issuer: "acme_tools_com",
           callbackUrl: "https://relyingparty/adfs/postResponse",
           privateCert: fs.readFileSync(__dirname + "/static/acme_tools_com.key", "utf-8"),
           authnContext:
             "http://schemas.microsoft.com/ws/2008/06/identity/authenticationmethod/password",
-          signatureAlgorithm: "sha256",
+          signatureAlgorithm: "sha256" as const,
```
cjbarth, line 2239: You shouldn't need `as const` here.

test/tests.ts
```diff
@@ -2167,86 +2229,90 @@ describe("passport-saml /", function () {
...
-        var samlConfig = {
+        const samlConfig = {
           entryPoint: "",
           issuer: "acme_tools_com",
           callbackUrl: "https://relyingparty/adfs/postResponse",
           privateKey: fs.readFileSync(__dirname + "/static/acme_tools_com.key", "utf-8"),
           authnContext:
             "http://schemas.microsoft.com/ws/2008/06/identity/authenticationmethod/password",
-          signatureAlgorithm: "sha256",
+          signatureAlgorithm: "sha256" as const,
```
cjbarth, line 2271: You shouldn't need `as const` here.

test/tests.ts
```diff
@@ -2167,86 +2229,90 @@ describe("passport-saml /", function () {
...
+        const samlConfig = {
           entryPoint: "https://adfs.acme_tools.com/adfs/ls/",
           issuer: "acme_tools_com",
           callbackUrl: "https://relyingparty/adfs/postResponse",
           privateCert: fs.readFileSync(__dirname + "/static/acme_tools_com.key", "utf-8"),
           authnContext:
             "http://schemas.microsoft.com/ws/2008/06/identity/authenticationmethod/password",
           identifierFormat: null,
-          signatureAlgorithm: "sha1",
+          signatureAlgorithm: "sha1" as const,
```
cjbarth, line 2304: You shouldn't need `as const` here.

test/tests.ts
```diff
@@ -2526,13 +2604,15 @@ describe("passport-saml /", function () {
...
+        const samlObjBadComparisonType = new SAML({ RACComparison: "bad_value" as any });
+        should.equal(
+          samlObjBadComparisonType.options.RACComparison,
+          "exact",
+          "the default value of the option `RACComparison` must be exact"
+        );
 
-        var validComparisonTypes = ["exact", "minimum", "maximum", "better"],
-          samlObjValidComparisonType;
+        const validComparisonTypes = ["exact", "minimum", "maximum", "better"] as const;
```
cjbarth, line 2614: This should be a type to itself. I'll push up a change to address this.

cjbarth (changes requested):
I found some simple things that need adjustment. I've also prepared some changes and pushed them to clean up the testing framework.

test/multiSamlStrategy.ts
```diff
@@ -101,11 +101,11 @@ describe("strategy#authenticate", function () {
       }
     }
 
-    var strategy = new MultiSamlStrategy(
-      { getSamlOptions: getSamlOptions, cacheProvider: "mock cache provider" },
+    const strategy = new MultiSamlStrategy(
+      { getSamlOptions: getSamlOptions, cacheProvider: "mock cache provider" as any },
```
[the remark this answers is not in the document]
gugu: the idea of the test is to violate typings and check that the user receives an error even if he has no typescript

test/multiSamlStrategy.ts
```diff
@@ -85,10 +85,10 @@ describe("strategy#authenticate", function () {
       identifierFormat: "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
       path: "/saml/callback",
       logoutUrl: "http://foo.slo",
-      signatureAlgorithm: "sha256",
+      signatureAlgorithm: "sha256" as const,
```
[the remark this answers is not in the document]
gugu: we need it, because if we don't add it here, it sets type as string and string is too broad for type 'sha1' | 'sha256‘ | ... used in the module

test/tests.ts
gugu: we need it, because if we don't add it here, it sets type as string and string is too broad for type 'sha1' | 'sha256‘ | ... used in the module

test/tests.ts
gugu: we need it, because if we don't add it here, it sets type as string and string is too broad for type 'sha1' | 'sha256‘ | ... used in the module

test/tests.ts
gugu: we need it, because if we don't add it here, it sets type as string and string is too broad for type 'sha1' | 'sha256‘ | ... used in the module

cjbarth: I understand what you are saying now... in that case, you can change it from `as const` to `as SignatureAlgorithm`. In fact, I've just pushed that change up.

test/samlTests.ts
```diff
@@ -67,7 +75,7 @@ describe("SAML.js", function () {
       it("calls callback with original query string", function (done) {
         saml.getAuthorizeUrl(req, {}, function (err, target) {
```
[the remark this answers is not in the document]
gugu: we need to rewrite the test to use getAuthorizeUrlAsync, but I suggest doing this in a separate PR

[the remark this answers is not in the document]
cjbarth: Ok. This makes sense. We can get this landed and then make another PR where we switch tests from the callback to the `async/await` pattern and remove all that deprecated code before we ship v3.

test/test-signatures.ts
```diff
@@ -27,21 +31,23 @@ describe("Signatures", function () {
         //== Run the test in `func`
         samlObj.validatePostResponse(
```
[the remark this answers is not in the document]
cjbarth: We'll address this in another PR.

test/multiSamlStrategy.ts
```diff
@@ -12,13 +12,13 @@ describe("Strategy()", function () {
...
     }
-    var strategy = new MultiSamlStrategy({ getSamlOptions: getSamlOptions }, verify);
+    const strategy = new MultiSamlStrategy({ getSamlOptions: getSamlOptions }, verify);
     strategy.should.be.an.instanceOf(SamlStrategy);
   });
 
   it("throws if wrong finder is provided", function () {
     function createStrategy() {
-      return new MultiSamlStrategy({}, verify);
+      return new MultiSamlStrategy({} as any, verify);
```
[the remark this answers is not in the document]
cjbarth: Addressed.

test/multiSamlStrategy.ts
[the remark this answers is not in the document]
cjbarth: Did some Googling, and it appears there isn't a great way to do this with current versions of TypeScript; though there are proposals to 'fix' this.

cjbarth:
I've merged in the head of `master`, but didn't get a chance to clean everything up.

test/tests.ts
gugu: I think strategy typings need major rewrite, I don't think they need impede our tests

cjbarth: I fixed this and will try to push something up later today.

cjbarth: I fixed this by using the types that we have defined in our project.

cjbarth:
There are some tests that fail intermittently, I think because we have a test-interdependence. So, when Mocha runs the tests in random order, some tests occasionally fail. Other than that, I think this code is all better. I would love someone to take a look at the intermittent failures.

gugu:
@cjbarth done. PassReqToCallback test modified shared `config` variable

cjbarth (approved):
This all looks good. I think we should probably make a follow-on PR that removes all the deprecated code and converts tests that use deprecated code to use the newer code.

If you feel it looks good, either one of us can do a merge commit (though I would clean up the comment because it is a little crazy now :) ).

Issue #533, opened 2021-02-09
Is it possible to get "WantAssertionsSigned = true" to work?


HendrikJan:
Our IdP is requiring that we set `WantAssertionsSigned` to true in the metadata we send them.

I did not find an option in the documentation to add `WantAssertionsSigned="true"` to the metadata when calling `generateServiceProviderMetadata( decryptionCert, signingCert )`.
I did not find the string "WantAssertionsSigned" anywhere in the code, so I assume this is not implemented.

Therefore I added it to the generated metadata this way:

```js
xml2js.parseString(metadata, (err, result) => {
        // Add WantAssertionsSigned to the metadata (as requested by IdP)
        result.EntityDescriptor.SPSSODescriptor[0].$.WantAssertionsSigned = 'true';

       // Convert JSON back to XML
      const builder = new xml2js.Builder();
      xml = builder.buildObject(result);
}
```

I was hoping the assertions would be signed anyway and thus this would work out fine.
However, the IdP now returns "NotAuthorized", and testing with SimpleSamlPhp as IdP, we now get the error: "Unable to validate signature on query string."
Unfortunately I am not very experienced with SAML and so far not able to solve this myself.

Does the current version of passport-saml support signing assertions, and what configuration options should I set?

cjbarth:
Thank you for your feedback about this feature. I don't believe we support this feature, particularly if you've already searched the code like you have. We aren't really in a position to implement features for others, but we do try to help others add the features correct and include them in this project.

We typically start by asking that a link to the relevant part of the SAML spec be linked to. Then a test need to be written to validate that the feature currently doesn't work which asserts that it is working like it should after it is implemented.

HendrikJan:
Hi @cjbarth, thank you for the quick reply.
I did solve my problem. Just manually adding `WantAssertionsSigned="true"` to the metadata XML is enough.
The reason my IdP rejected was because I set encryption to sha512, which was not supported by the IdP.

In case anybody else is looking for "WantAssertionsSigned" and stumbling upon this issue:
The SAML Spec says this:

> WantAssertionsSigned [Optional]
[...]

PR #536, opened 2021-02-13
Add WantAssertionsSigned

HendrikJan:
# Description

This is related to #533 

This change adds the option "wantAssertionsSigned" to the SamlStrategy options.
When this option is set to true, `WantAssertionsSigned="true"` is added to the metadata when calling `samlStrategy.generateServiceProviderMetadata(...)`.
When this option is set to false, `WantAssertionsSigned` is not added to the metadata, which is de default.

This feature is described in the SAML Spec:

> WantAssertionsSigned [Optional]
> Optional attribute that indicates a requirement for the <saml:Assertion> elements received bythis service provider to be signed.If omitted, the value is assumed to be false. This requirementis in addition to any requirement for signing derived from the use of a particular profile/bindingcombination. [E7]Note that an enclosing signature at the SAML binding or protocol layer does notsuffice to meet this requirement, for example signing a <samlp:Response> containing theassertion(s) or a TLS connection.

(Line 828 of this document: [sstc-saml-metadata-errata-2.0-wd-04-diff.pdf](https://www.oasis-open.org/committees/download.php/35391/sstc-saml-metadata-errata-2.0-wd-04-diff.pdf) )

The use-case is that some IdP's require `WantAssertionsSigned="true"` in the metadata that you send them.

_Please include tests._ Doing so will ensure that the changes made in this PR are not undone or otherwise corrupted by future changes.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [x]
- Tests included? [x]
- Documentation updated? [x]


README.md
```diff
@@ -125,6 +125,7 @@ type Profile = {
 - `additionalParams`: dictionary of additional query params to add to all requests; if an object with this key is passed to `authenticate`, the dictionary of additional query params will be appended to those present on the returned URL, overriding any specified by initialization options' additional parameters (`additionalParams`, `additionalAuthorizeParams`, and `additionalLogoutParams`)
 - `additionalAuthorizeParams`: dictionary of additional query params to add to 'authorize' requests
 - `identifierFormat`: if truthy, name identifier format to request from identity provider (default: `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`)
+- `wantAssertionsSigned`: it truthy, add `WantAssertionsSigned="true"` to the metadata, to specify that the IdP should always sign the assertions.
```
cjbarth, line 128: You probably meant "if truthy" instead of "it truthy".

cjbarth (changes requested):
A small change needed for the readme, otherwise this looks good. Thank you for your contribution!

srd90:
Prior to this pull request passport-saml's certificate check seems to work like this:

1. If `cert` configuration option is not present at all response's and assertions' signature are not checked at all (and message can be altered in any way before submitted to passport-saml's assertion consuming service endpoint)
2. If `cert` contains certificate or array of certificates (i.e. value of `cert` is truthy) passport-saml checks whether response has been signed:
https://github.com/node-saml/passport-saml/blob/54704deb26b3cd495882ef9d12a97c1130e2b64f/src/passport-saml/saml.ts#L849-L852
and sets `validSignature = true` if reponse's signature was valid.
If response's signature was valid passport-saml **skips** assertion's signature:
https://github.com/node-saml/passport-saml/blob/54704deb26b3cd495882ef9d12a97c1130e2b64f/src/passport-saml/saml.ts#L870-L877
and
https://github.com/node-saml/passport-saml/blob/54704deb26b3cd495882ef9d12a97c1130e2b64f/src/passport-saml/saml.ts#L902-L909
i.e. passport-saml is happy if response's signature was valid and it does not try to validate assertion's signature at all.

IMHO `WantAssertionsSigned = true` at SAML SP metadata means what referenced spec says:
>Note that an enclosing signature at the SAML binding or protocol layer does not suffice to meet this requirement, for example signing a `<samlp:Response>` containing the assertion(s) or a TLS connection.

i.e. that even if response was signed and even if response's signature was valid SP expects that assertion is also signed and that SP shall verify assertion's signature before consuming it.

It looks like this pull request adds the possibility to set`WantAssertionsSigned = true` to medata generated by passport-saml just to satisfy some IdP administrator(s) requirements that SP connecting to their IdP must validate signatures of assertions or not use their IdP at all. I.e. passport-saml's internal implementation of `WantAssertionsSigned = true` case is missing and it - passport-saml - would not work the way its metadata implicates.

cjbarth:
@srd90 , you make a good point. So it seems then for this PR to be truly compliant, it should actually perform more validations instead of just saying it does. @HendrikJan , are you able to add this additional validation. @srd90 any ideas how we might test that we performed the check?

HendrikJan:
@srd90 , that makes a lot of sense. Thank you.
@cjbarth. I will look into this later this week or in the weekend. With help of the comments already given by @srd90 I may be able to pull this off.

HendrikJan:
I've updated the pieces of code where @srd90 noticed the assertions' validation being skipped.

Testing with SimpleSamlPHP confirms that we now get the error message "Invalid signature" when the IdP does not sign the assertions.

I did not yet think about how to test this.

    Issue Addressed: [x]
    Link to SAML spec: [x]
    Tests included? [ ]
    Documentation updated? [x]

test/tests.js
```diff
@@ -1660,6 +1660,24 @@ describe("passport-saml /", function () {
       metadata.should.containEql(samlConfig.logoutCallbackUrl);
     });
 
+    it("generateServiceProviderMetadata contains WantAssertionsSigned", function () {
```
cjbarth, line 1663: We will also need a test to verify that a bad assertion signature is caught.

cjbarth (changes requested):

srd90:
There seems to be at least following scenarios which should be covered with new tests when passport-saml stack is configured with `WantAssertionsSigned = true`:

1. signed response with unsigned unencrypted assertion
    * expected result is invalid signature
2. signed response with unsigned encrypted assertion
    * because program logic ("if" statement which determines when assetion signature should be checked) is and has been duplicated to code blocks which handle encrypted and unencrypted assertions
    * expected result is invalid signature
3. unsigned response with signed unencrypted assertion and `cert` option not provided at all
    * because passport-saml can be configured without `cert` by not defining that configuration attribute at all (see issues 180 and 523) in which case passport-saml shall allow anyone to manipulate login response. This would be against `WantAssertionsSigned = true` and passport-saml should assume that desired behaviour is validly signed assertion (i.e. if passport-saml implementation is not modified to fail during stack setup phase due conflicting configuration parameters assertion validation should fail at least during validation of signature by XML signature library due undefined `cert`)
    * expected result is invalid signature
4. unsigned response with with signed encrypted assertion and `cert`option not provided at all (see 3.)
    * reason behind this is duplicated program logic (aforementioned "if" statement)
    * expected result invalid signature
5. signed response with signed unencrypted assertion so that assertions content (e.g. nameid) is modified after calculation of assertion's signature but prior to signing response (i.e. response's signature is valid but assertion's signature would not be)
    * expected result invalid signature
6. signed response with signed encrypted assertion so that assertion's content doesn't match with assertion's signature
     * i.e. same as 5. but for encrypted assertion (e.g. by modifying assertion's nameid after assertion's signature is calculated but prior to encrypting assertion and signing response)
     * expected result invalid signature

HendrikJan:
@srd90 thanks for thinking about all those scenarios.
I'll try to get at least some of this to work this weekend.

HendrikJan:
@cjbarth and @srd90 

I've added test 1 and 3 and an extra test to catch a missing cert when initializing saml.

Now this branch cannot merge anymore. How should I proceed to fix the merge-conflict?
- Can I merge with the current master and fix conflicts
- Or should I rebase (I tried that, but I'm not experience with rebasing and after the rebase I got the error "Failed at the passport-saml@2.0.5 lint script." and I don't know how to fix this.)

(The rebased branch can be found here: https://github.com/HendrikJan/passport-saml/tree/wantassertionssigned_rebase)

cjbarth:
@HendrikJan I fixed the conflicts for you, but there is still some work to do to pass the tests.

HendrikJan:
@cjbarth Thank you. I know there are still tests to be made and to be fixed.

HendrikJan:
@cjbarth and @srd90

I've added all tests suggested by @srd90.

I've created the new encrypted and signed responses using the Onelogin saml tool (https://www.samltool.com/encrypt.php). I hope that is an accepted way to do it.

test/test-signatures.spec.ts
```diff
@@ -86,6 +95,64 @@ describe("Signatures", function () {
...
     );
+    it(
+      "R1A - root signed - wantAssertionsSigned=true => error",
+      testOneResponse("/valid/response.root-signed.assertion-unsigned.xml", INVALID_SIGNATURE, 2, {
+        cert,
+        wantAssertionsSigned: true,
+      })
+    );
+    it(
+      "R1A - root unsigned - missing cert - wantAssertionsSigned=true => error",
```
srd90, line 106: IMHO this test case and `R1A - root unsigned - asrt signed encrypted - missing cert - wantAssertionsSigned=true => error` should be augmented with code comment which states that these are "absurd" test cases because `passport-saml` should never allow situation where it consumes unsigned authentication response so that assertion is also unsigned (regardless of `WantAssertionsSigned` configuration option's value) which is exactly what `passport-saml` does when `cert` option is completely missing (or misspelled). I.e. these two cases exists just to test that at least `WantAssertionsSigned = true` situation fall to safe direction (i.e. reporting error instead of consuming  assertion).

This and `R1A - root unsigned - asrt signed encrypted - missing cert - wantAssertionsSigned=true => error` cases could be removed when `cert` is made truly mandatory configuration option ( see issues 180 and 523) and existence of `cert` option is validated during `passport-saml` setup phase.

With aforementioned code comment in place someone in the future would not have to spend time wondering whether he/she should fix these two cases somehow after `cert` is changed to mandatory value and these cases start to fail during test setup phase.

Sidenote regarding "absurdness" of allowing missing `cert` configuration and how `passport-saml` works when it doesn't have that option: For example Shibboleth IdP's documentation has following note (especially second paragraph):
>The default value of signResponses for this profile is "true", in keeping with modern best practice. As long as one of the response or assertion are signed, use of the profile is "safe" in terms of authentication integrity, but there are vulnerabilities in XML Encryption that make signing responses advisable when the most common encryption algorithms are used. Some of the backstory around the signing defaults is discussed in this thread.
>
>If you encounter a relying party that accepts an unsigned response **and** assertion that is transmitted via POST (and not artifact), you have identified an insecure implementation and should report the issue immediately while following your local security incident response process.

source: https://wiki.shibboleth.net/confluence/display/IDP4/SAML2SSOConfiguration
(exact version of that particular wiki page: https://wiki.shibboleth.net/confluence/pages/viewpage.action?pageId=55804373 )

src/passport-saml/saml.ts
```diff
@@ -1554,6 +1556,13 @@ class SAML {
       metadata.EntityDescriptor.SPSSODescriptor.NameIDFormat = this.options.identifierFormat;
     }
 
+    if (this.options.wantAssertionsSigned) {
+      if (!this.options.cert) {
+        throw new Error('"cert" config parameter is required for signed assertions');
```
srd90, line 1561: I think this this option combination check should be at SAML's constructor/initialize method. Now this check is evaluated if and only if `generateServiceProviderMetadata` method is invoked. Earlier comments about allowing missing `cert` at all (regardless of value of `wantAssertionsSigned`) apply.

If this option combinarion check is moved to constructor/initialize method then you would not have to have this at `validatePostResponseAsync` method:
```javascript
      if (this.options.wantAssertionsSigned && !this.options.cert) {
        throw new Error("Invalid signature");
      }
  ```
because one would have had error during constructor invocation (if I interpreted javascript/typescript correctly...I'm not too familiar with these languages).

Comment (HendrikJan) [reply src/passport-saml/saml.ts:1561]: Thank you for the constructive feedback.
Moving the option combination check might be an improvement indeed; I'll look into it.

File: test/test-signatures.spec.ts (hunk 3 above)
Comment (HendrikJan) [reply test/test-signatures.spec.ts]: I've removed the wantAssertionsSigned+cert combination check to the saml initialization.
Now these two "absurd" testcases are not needed anymore because passport-saml throws at initialization.

I've removed the tests.

File: src/passport-saml/saml.ts (hunk 4 above)
Comment (HendrikJan) [reply src/passport-saml/saml.ts]: I've updated the code as you suggested.

File: test/tests.js (hunk 2 above)
Comment (HendrikJan) [reply test/tests.js]: Bad assertion signature is now tested in these tests:

https://github.com/node-saml/passport-saml/blob/f6bce345fe6d7af58a71d246290ae9d1b2167227/test/test-signatures.spec.ts#L118-L119

https://github.com/node-saml/passport-saml/blob/f6bce345fe6d7af58a71d246290ae9d1b2167227/test/test-signatures.spec.ts#L130-L131

**Review by cjbarth — APPROVED** [review]
This looks good to me, but @srd90 has spent more time with this than I have, so I'd like him to sign off on it before this gets merged.

File: src/passport-saml/saml.ts [hunk 5]
```diff
@@ -131,6 +131,10 @@ class SAML {
       throw new Error("Invalid property: cert must not be empty");
     }
 
+    if (options.wantAssertionsSigned && !options.cert) {
+      throw new Error('"cert" config parameter is required for signed assertions');
```
srd90, line 135: This error message looks a bit misleading. I suggest: "cert config parameter with IdP's certificate is required to validate assertion's signature".

side note: @cjbarth @markstos IMHO this if statement and the one at the line 130 of this pull request branch could be combined into single block which checks that cert configuration option is defined and that value of cert is not undefined (that check would cover also this particular blocks situation because it would make sure that cert is provided). Now these two if statements are used to keep ignoring missing cert option situation and result is issues 180 and 523

srd90:
@cjbarth regarding this comment:
>...I'd like him to sign off on it before this gets merged

I haven't executed single line of new code introduced in this pull request.
Comments were based only on reading the code (and I'm not javascript/typescript coder so my internal interpreter might work incorrectly). But:

It seems that this pull request's initial change (`WantAssertionsSigned = true` at generated SP metadata) is now (at the end of this pull request) respected also by `passport-saml` itself.

Furthermore as a result of this pull request few "if" statements related to certifcate checks are a bit more complicated but it seems that those "if" statements works as before (i.e. if `wantAssertionsSigned = false` and `cert` is defined then validly signed assertion is sufficient if validation of response failed and if `wantAssertionsSigned = false`and `cert` option is missing then there is no security at all).

About sign off: Sorry, I'm not going to sign off this pull request because as a whole certificate checking is broken (it is not a fault of this pull request). Its up to maintainers to sign off/accept next version of certificate checking with known issues.

btw. @HendrikJan this pull request's feature is implemented in 37 separated commit are you going to ship all those commits to passport-saml repository instead of squashing to one commit before merge(?)

test/test-signatures.spec.ts
```diff
@@ -86,6 +95,51 @@ describe("Signatures", function () {
...
+        2,
+        {
+          decryptionPvk: fs.readFileSync(__dirname + "/static/testshib encryption pvk.pem"),
+          cert,
+          wantAssertionsSigned: true,
+        }
+      )
+    );
+    it(
+      "R1A - root signed - asrt invalidly signed wantAssertionsSigned=true => error",
```
srd90, line 119: @HendrikJan I tried to run this testcase in debugger in order to verify that it works as expeced i.e. that:
>5. signed response with signed unencrypted assertion so that assertions content (e.g. nameid) is modified after calculation of assertion's signature but prior to signing response (i.e. response's signature is valid but assertion's signature would not be)
> * expected result invalid signature

Purpose was that response's signature is valid and this code block at saml.ts:
https://github.com/node-saml/passport-saml/blob/bb0c32d0bf2532fdb706e5017495bff8674fdf91/src/passport-saml/saml.ts#L857
would set `validSignature = true` because top / msg level signature matched content.

Now - when `validSignature = true` due valid response level signature - assertion's signature check (because `wantAssertionsSigned = true`) should fail because `passport-saml` is instructed to check also assertion's signature even if top level signature was considered valid and overall result should be invalid signature exception.

Well...this test case ends up checking assertion's signature but due wrong reason (result would have been same even if `wantAssertionsSigned` would have been `false` because `passport-saml` tries to check assertions signature if saml responses signature was invalid).

Purpose of this test case was to to make sure/test that `passport-saml` checks assertions signature (when `wantAssertionsSigned = true`) EVEN if `validSignature` is already set to `true` due valid response level signature. In order to accomplish this test I suggested "validly signed saml reposponse which contains assertion with invalid assertion signature" because normally (when `wantAssertionsSigned = false`) such message would be considered valid (assertion's signature checking would be skipped).

It was quite painful exercise to setup mocha typescript testcase debugging env (without any proper js/ts toolchain background) so I might have done something incorrectly  but result (after stepping through `... mocha ... -g "R1A - root signed - asrt invalidly signed wantAssertionsSigned=true => error" ...`) seems to be that `invalid/response.root-signed.assertion-invalidly-signed.xml` is not constructed correctly because `passport-saml` considered top level signature to be invalid (it should have been valid so that test case works as expected). I assume `invalid/response.root-signed.assertion-invalidly-signed-encrypted.xml` has similar problem.

HendrikJan:
@srd90
Regarding your remark on squishing al those 37 commits into one.
I would love to combine them into one commit.
@cjbarth Is that an option once this PR is (hopefully) approved?

test/test-signatures.spec.ts
HendrikJan: Thank you so much for taking all this time and even setting up a debugging environment.

This PR has become much more involved than my original (very naive as I can see now) pull request.
Working with XML and signing and decrypting is new for me so I could easily have made a mistake.
I'll try to set up debugging later this week and step through the code.

markstos:
@srd90 I'm arriving here late and may have missed some context. 

First, thanks for your help reviewing this. When you say the "whole certificate checking is broken", could you provide more context or a link to a reference about this?

cjbarth:
> 
> 
> @srd90
> Regarding your remark on squishing al those 37 commits into one.
> I would love to combine them into one commit.
> @cjbarth Is that an option once this PR is (hopefully) approved?

We can do a squash-merge when the code lands, so don't worry about that.

cjbarth:
> First, thanks for your help reviewing this. When you say the "whole certificate checking is broken", could you provide more context or a link to a reference about this?

@markstos I believe it is already well explained in #523 .

HendrikJan:
I could use some help here.
I am missing the knowledge to generate saml-responses for the tests.

I tried to use https://www.samltool.com to sign and encrypt saml-responses, but everything I generate fails at the wrong place.

Does anybody have a suggestion on how to generate the following saml responses:

1. signed response with unsigned encrypted assertion 
2. unsigned response with with signed encrypted assertion
3. signed response with signed unencrypted assertion so that assertions content (e.g. nameid) is modified after calculation of assertion's signature but prior to signing response
4. signed response with signed encrypted assertion so that assertion's content doesn't match with assertion's signature

test/test-signatures.spec.ts
HendrikJan: I've updated the response xml files.

HendrikJan:
@cjbarth and @srd90 

I've updated the test.

In the following piece of code:

      // Check if this document has a valid top-level signature
      let validSignature = false;
      if (this.options.cert && this.validateSignature(xml, doc.documentElement, certs!)) {
        validSignature = true;
      }

The line `validSignature = true;` is now run for every test and this seems correct to me because all tests have a signed root which should be valid.

Also all tests have `amountOfSignaturesChecks = 2`. One check is for the root and one for the assertion which should always be checked if `wantAssertionsSigned = true`

cjbarth:
@HendrikJan This looks pretty good to me. I have a PR, #548 , that I'm eger to land first since it addresses a security concern, and would actually force changes in this PR. I wouldn't mind helping with merge conflicts if that is needed, but I'll hold off on approving this until I can manage to get the unexpectedly complicated #548 landed.

I'm also interested to see if @srd90 has anything to say about your latest changes or what is in #548 which addresses the concerns that you and he raised in #523.

cjbarth:
@HendrikJan Would you like to do your rebase now so we can get this landed as part of the upcoming 3.x release? If you need help, please let me know.

HendrikJan:
@cjbarth I'll have a look this weekend.

HendrikJan:
@cjbarth I've resolved the merge conflict.

README.md
```diff
@@ -126,6 +126,7 @@ type Profile = {
 - `additionalParams`: dictionary of additional query params to add to all requests; if an object with this key is passed to `authenticate`, the dictionary of additional query params will be appended to those present on the returned URL, overriding any specified by initialization options' additional parameters (`additionalParams`, `additionalAuthorizeParams`, and `additionalLogoutParams`)
 - `additionalAuthorizeParams`: dictionary of additional query params to add to 'authorize' requests
 - `identifierFormat`: optional name identifier format to request from identity provider (default: `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`)
+- `wantAssertionsSigned`: if truthy, add `WantAssertionsSigned="true"` to the metadata, to specify that the IdP should always sign the assertions.
```
cjbarth, line 129: Actually, "truthy" isn't the correct thing to say here. The `??` will test for ` == null`, which is `null` and `undefined`. We might want to change the ctor to coerce this value to boolean using `!!` if that is what we really want.

src/passport-saml/saml.ts
```diff
@@ -147,6 +147,7 @@ class SAML {
         ctorOptions.identifierFormat === undefined
           ? "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"
           : ctorOptions.identifierFormat,
+      wantAssertionsSigned: ctorOptions.wantAssertionsSigned ?? false,
```
cjbarth, line 150: If this is supposed to be boolean, then we should probably force a coercion here with `!!`.

test/test-signatures.spec.ts
```diff
@@ -1,22 +1,25 @@
...
+    INVALID_ENCRYPTED_SIGNATURE = "Invalid signature from encrypted assertion",
     createBody = (pathToXml: string) => ({
       SAMLResponse: fs.readFileSync(__dirname + "/static/signatures" + pathToXml, "base64"),
     }),
     testOneResponseBody = async (
       samlResponseBody: Record<string, string>,
       shouldErrorWith: string | false | undefined,
-      amountOfSignatureChecks = 1
+      amountOfSignatureChecks = 1,
+      options: Partial<SamlOptions> = {}
```
cjbarth, line 19: Why not use the type `SamlConfig` instead as that better represents what a consumer should be using.

test/test-signatures.spec.ts
```diff
@@ -30,11 +33,17 @@ describe("Signatures", function () {
     testOneResponse = (
       pathToXml: string,
       shouldErrorWith: string | false,
-      amountOfSignaturesChecks: number | undefined
+      amountOfSignaturesChecks: number | undefined,
+      options?: Partial<SamlOptions>
```
cjbarth, line 37: Why not use the type SamlConfig instead as that better represents what a consumer should be using.

cjbarth (changes requested):
Some minor things, but the overall code looks good. Thanks for be patient and cooperative with getting this landed.

README.md
HendrikJan: If we use `!!`, that would be the odd one out within all the other optional parameters that use `?? false`.

A list of all options using `?? false` and the wording in the readme:
* passive: **not found in readme**
* disableRequestedAuthnContext: "if truthy"
* forceAuthn: "if set to true"
* skipRequestCompression: "if set to true"
* disableRequestAcsUrl: "if truthy"
* wantAssertionsSigned: "if truthy"
* validateInResponseTo: "if truthy"

By using the phrase "if truthy" I follow the same structure as most of the options.

I would prefer to use the same wording everywhere, and to me "if set to true" sounds more professional and has my preference, but than we should change all these options in the readme.

These changes in the readme should probably be done in another PR.
Probably combined with alphabetically ordering the readme and the code, and adding `passive` to the readme.

src/passport-saml/saml.ts
HendrikJan: I think the changes should be done in `README.md` in another PR.

test/test-signatures.spec.ts
HendrikJan: Good suggestion. I've updated the code.

test/test-signatures.spec.ts
HendrikJan: Good suggestion. I've updated the code.

cjbarth (approved):

PR #545, opened 2021-02-23
remove old callback functions, tests use async/await

gugu:


test/tests.spec.ts
```diff
@@ -2698,16 +2606,12 @@ describe("passport-saml /", function () {
...
-            should.exist(err);
-            err!.message!.should.match("InResponseTo is not valid");
-            done();
-          } catch (err2) {
-            done(err2);
-          }
-        });
+        try {
+          await samlObj.validatePostResponseAsync(container);
+          false.should.eql(true, "Should not succeed");
```
cjbarth, line 2611: Isn't there an assert that says something should throw?

test/tests.spec.ts
```diff
@@ -2979,19 +2865,15 @@ describe("passport-saml /", function () {
...
-            should.exist(err);
-            err!.message!.should.match("SAML assertion expired");
-            done();
-          } catch (err2) {
-            done(err2);
-          }
-        });
+        try {
+          await samlObj.validatePostResponseAsync(container);
+          false.should.eql(true, "Should not succeed");
```
cjbarth, line 2870: Is there a reason we have to use this pattern instead of https://shouldjs.github.io/#should-throws?

test/tests.spec.ts
```diff
@@ -3142,82 +2998,63 @@ describe("passport-saml /", function () {
...
-          should.exist(err);
-          err!.should.eql(new Error("Invalid signature on documentElement"));
-          done();
-        } catch (err2) {
-          done(err2);
-        }
-      });
+      try {
+        await samlObj.validatePostRequestAsync(body);
+        false.should.eql(true, "validatoPostRequest should fail");
```
cjbarth, line 3021: Spelling mistake.

test/tests.spec.ts
```diff
@@ -3256,296 +3086,256 @@ describe("passport-saml /", function () {
...
-          JSON.parse(fs.readFileSync(__dirname + "/static/sp_slo_redirect.json", "utf8"))
+    afterEach(function () {
+      this.clock.restore();
+    });
+    it("errors if bad xml", async function () {
+      const body = {
+        SAMLRequest: "asdf",
+      };
+      try {
+        await samlObj.validateRedirectAsync(body, null);
```
cjbarth, line 3200: This test changed slightly. It used to include a valid redirect URL, but now just has `null`.

test/tests.spec.ts
```diff
@@ -3256,296 +3086,256 @@ describe("passport-saml /", function () {
...
-            should.exist(err);
-            done();
-          } catch (err2) {
-            done(err2);
-          }
-        });
+      }
+    });
+    it("errors if request has expired", async function () {
+      this.clock.restore();
```
cjbarth, line 3218: This shouldn't be needed if we are resetting the clock at the end of other tests.

test/tests.spec.ts
```diff
@@ -3256,296 +3086,256 @@ describe("passport-saml /", function () {
...
+    });
+    it("errors if bad xml", async function () {
+      const body = {
+        SAMLRequest: "asdf",
+      };
+      try {
+        await samlObj.validateRedirectAsync(body, null);
+        false.should.equal(true, "validateRedirectAsync should fail");
+      } catch (err) {
+        // as expected
```
cjbarth, line 3275: We should also use the proper asserts for this instead of this pattern.

cjbarth (changes requested):
Overall, pretty good. There was just one spelling mistake I found and a few places where it seems we are using the wrong assert.

test/tests.spec.ts
cjbarth: `import assert = require("assert");`
`assert.doesNotReject(samlObj.validatePostResponseAsync(container));`

cjbarth: `assert.rejects(samlObj.validatePostResponseAsync(container), /Invalid signature/);`

test/tests.spec.ts
gugu: it does not work for async functions. Should.js is a deprecated module

cjbarth: You can use the pattern I highlighted above instead:

```js
import assert = require("assert");
assert.rejects(samlObj.validatePostResponseAsync(container), /Invalid signature/);
```

test/tests.spec.ts
gugu: looks like this test needs real date, that's why it resets clock before

cjbarth (approved):

cjbarth:
Don't forget to squash merge.

PR #546, opened 2021-02-24
Messed up with a valid response and it still tests as valid.

HendrikJan:
# Description

I am unsure if this new test shows a bug in the response validation.

For this test I copied the valid response `valid/response.root-signed.assertion-signed.xml` and replaced the text "https://evil-corp.com" for "https://hacker-corp.com".
I placed this invalid response in `invalid/response.root-invalidly-signed.assertion-signed.xml`.

This new file should be an invalid response (because I have tempered with the contents).
Therefore the test should result in "Invalid signature", but what it does return is: "SAML assertion expired".

Is this an error, or am I missing something here?

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [x]
- Documentation updated? [ ]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 yarn-error.log
 .DS_Store 
 .eslintcache
+.vscode
--- a/test/static/signatures/invalid/response.root-invalidly-signed.assertion-signed.xml
+++ b/test/static/signatures/invalid/response.root-invalidly-signed.assertion-signed.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" Destination="https://evil-corp.madness.com/sso/callback" ID="_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" InResponseTo="_e8df3fe5f04237d25670" IssueInstant="2015-08-31T08:54:06+00:00" Version="2.0">
+    <saml:Issuer>https://hacker-corp.com</saml:Issuer>
+    <samlp:Status>
+        <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
+    </samlp:Status>
+    <saml:Assertion ID="_bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" IssueInstant="2020-09-25T16:00:00+00:00" Version="2.0">
+        <saml:Issuer>https://evil-corp.com</saml:Issuer>
+        <saml:Subject>
+            <saml:NameID Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress">vincent.vega@evil-corp.com
+            </saml:NameID>
+            <saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
+                <saml:SubjectConfirmationData InResponseTo="_e8df3fe5f04237d25670" NotOnOrAfter="2020-09-25T16=7:00:00+00:00" Recipient="https://evil-corp.madness.com/sso/callback"/>
+            </saml:SubjectConfirmation>
+        </saml:Subject>
+        <saml:Conditions NotBefore="2020-09-25T16:00:00+00:00" NotOnOrAfter="2020-09-25T17:00:00+00:00"/>
+        <saml:AuthnStatement AuthnInstant="2020-09-25T16:00:00+00:00" SessionIndex="_9e315bdf7b1b6732be33c377cf6f5c4f">
+            <saml:AuthnContext>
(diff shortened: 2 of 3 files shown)


srd90:
First of all lets inspect how test case which uses  `valid/response.root-signed.assertion-signed.xml` behaves because it is quite odd that your test receives "Assertion expired" (one could have expected for example issuer mismatch if/when issuer element at samlp namespace mismatch with issuer element at saml namespace).
So  when `R1A - both signed => valid` test case is executed (link points to version at passport-saml's master branch):
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/test/test-signatures.spec.ts#L50-L53
I would not expect to see assertion expired because I would expect that test case controls current time behind the scenes so that response is processed without any exceptions.

Lets put break point to both of these lines and see what happens when aforementioned test case is executed:
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/src/passport-saml/saml.ts#L1262-L1263
First time when we hit break points value of `notOnOrAfter` is `2020-09-25T16=7:00:00+00:00` (note `=` character) because of this
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/test/static/signatures/valid/response.root-signed.assertion-signed.xml#L13
and `notOnOrAfterMs` becomes `NaN` and test is `some value >= NaN` ... why author of `response.root-signed.assertion-signed.xml` and this testase wanted to have such value...dunno. Should there be failure due this issue ... don't have time to investigate...

Second  time break points are hit when this
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/test/static/signatures/valid/response.root-signed.assertion-signed.xml#L16
is processed. This time code ends up executing this:
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/src/passport-saml/saml.ts#L1263

So `R1A - both signed => valid` is having same "Assertion expired" issue under the hood. Why is test case passing? Dunno. Maybe someone who really knows what happens at these lines can explain why this test case is reporting success:
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/test/test-signatures.spec.ts#L32-L36
because message processing is definitely throwing exception while at the same time it is invoking signature check only once. Its as if passing `validateSignatureSpy.callCount.should.eql(amountOfSignatureChecks);` would be considered sufficient even if message processing ended up having exception.

Eitherway reference testcase seems to have issues/problems of its own. From your pull request's/question's point of view "Assertion expired" is not a "surprise".

---

Now lets move to case you have in this pull request.

passport-saml considers missing or invalid top level signature as "soft error" (assuming that `cert` option with IdP certificate was provided).
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/src/passport-saml/saml.ts#L783-L786
i.e. it leaves value of `validSignature` to `false` and proceeds to check whether assertion's signature is valid:
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/src/passport-saml/saml.ts#L804-L811
and because you modified only content of surrounding `samlp` stuff instead of assertion's stuff signature of assertion's content is valid and passport-saml proceeds to consume it. See also https://github.com/node-saml/passport-saml/pull/536#issuecomment-778838859

Quote from https://wiki.shibboleth.net/confluence/pages/viewpage.action?pageId=55804373
>As long as one of the response or assertion are signed, use of the profile is "safe" in terms of authentication integrity, but there are vulnerabilities in XML Encryption that make signing responses advisable when the most common encryption algorithms are used.

It would be a huge security problem if passport-saml would return any values from `samlp` namespace to application because those values MAY be modified by who ever is passing saml login response which - login response - has only/also validly signed assertion to passport-saml (because missing or invalid top level signature is considered "soft error" and if at least assertion signature is valid login is accepted). Quick and definitely **not too careful code browsing** gives an impression that values returned to application are taken from assertion and none of the values from `samlp` namespace are returned to application.

You would have to introduce another flag to passport saml so that it would consider missing or invalid top level signature as "hard error". Of course you would have to make sure that your IdP is signing samlp (response). For example Shibboleth IdP signs response by default and assertion signing is optional (if e.g. SP gives "WantAssertionsSigned = true").

Your pull request's testcase 
https://github.com/node-saml/passport-saml/blob/d6557d11bbfe6be05928c1f4b68c9849d447a1d6/test/test-signatures.spec.ts#L76-L83
is failing with the "Assertion expired" under the hoods due same reason as reference case examined at the start of this comment. Why your testcase is failing with such hard error while reference case is not ... dunno ... 

But at the end of the date and based on how passport-saml is currently implemented testcase you defined is actually "success / valid" case i.e. similar to
https://github.com/node-saml/passport-saml/blob/8a1a377f568fa017643404ba67127cd6928e1581/test/test-signatures.spec.ts#L50-L53
with the exception that processed response file has invalid top level signature and amount of signature checks would be 2 (and passport-saml would still throw "SAML assertion expired" but test case would be success due same strange reason as reference case is passing).

Issue #334, opened 2019-01-06
MultiSamlStrategy and InResponseTo are Incompatible by Default


gluxon:
For every HTTP request, `MultiSamlStrategy` [creates a brand new `SAML` object](https://github.com/bergie/passport-saml/blob/master/multiSamlStrategy.js#L24). This in turn [creates a new `InMemoryCacheProvider`](https://github.com/bergie/passport-saml/blob/master/lib/passport-saml/saml.js#L68).

Meaning by default, the `InResponseTo` setting will always cause the middleware to fail with *"InResponseTo is not valid"*. Cache keys will never persist beyond the life time of a single HTTP request unless a custom `cacheProvider` is supplied.

I don't think the usage of two officially supported parts of the library should silently break. There are a 2 ways to fix this (that I see):

1. Add an error if a custom `cacheProvider` isn't supplied by the user when using the `MultiSamlStrategy.` Forcing users to reconfigure this library with an external cache store would workaround this issue.
2. Have `MultiSamlStrategy` hold onto its own`InMemoryCacheProvider` objects by default that it passes to new `SAML` instances. To do this securely, it would need to persist caches for each identity provider.

I lean towards 2, but unfortunately the current API design doesn't require each identity provider to be uniquely keyed. The `endpoint` SAML option gets close, but that can change with binding methods.

**Edit**: To be clear, I'm just hoping to open a discussion on how to best resolve this. I'm not demanding it to be fixed or trying to force a solution. I may be able to submit a pull request once we agree on something.

gluxon:
I'm not sure how big of a problem it would be in practice, but it would allow a malicious identity provider to respond to an authentication request made to a different identity provider (assuming SAML response signing is off). If response signing is enforced, then not partitioning the cache should be fine.

@stavros-wb I just realized that you wrote MultiSamlStrategy. Thanks for that! It's been very useful to me.

stavros-wb:
@gluxon can you try with the latest master?

gluxon:
@stavros-wb Sorry for the late response, but this works! Thank you!

crossan007:
It seems like 3.2.0 no longer defaults to `InMemoryCacheProvider` for `MultiSamlStrategy`  (possibly as of 224f25f71156a5908fc214bd146ad6e8274a0a9f).

Any ideas if this was an intentional change?

It seems like I should still be able to explicitly define a `InMemoryCacheProvider`  in my initialization options as  `cacheProvider`

[...]

PR #548, opened 2021-02-26
Require cert for every strategy

### LINKED ISSUE #523: Missing cert configuration option disables signature validation silenty/without huge warning messages
There is/was issue #180 which addressed this particular case.

There is no need to repeat reasons why `passport-saml` MUST have safe defaults for signature validation configuration ( read @pitbulk 's and @alexstuart 's comments from #180 ). In fact one should not be able disable it at any circumstances. 

It seems that there was some sort of consensus at the end of #180 that
>Certificate must be required and validation for assertions must fail if signature is missing.

**but** content of commit ( f6b1c885c0717f1083c664345556b535f217c102 ) which closed issue #180 allows completely missing or misspelled `cert` configuration option which is effectively same thing as `cert` with `falsy` value (signature validation is completely disabled allowing anyone to alter login response any way he/she chooses to).

---

Following examples are executed against `passport-saml` version https://github.com/node-saml/passport-saml/tree/932da9d09a018fed4cb830e67090bb994f8539c1 ( `v2.0.4` ) using this particular test case (with various modifications) to demonstrate effects of missing `cert` configuration option:
https://github.com/node-saml/passport-saml/blob/932da9d09a018fed4cb830e67090bb994f8539c1/test/tests.js#L1623-L1643

---

**Reference run** (version 932da9d09a018fed4cb830e67090bb994f8539c1 without any modifications):
```plaintext
passport-saml$ npm run tsc && node_modules/mocha/bin/mocha -g 'valid onelogin xml document should validate'

  passport-saml /
    saml.js / 
      validatePostResponse checks /
        validatePostResponse xml signature checks /
          ✓ valid onelogin xml document should validate (81ms)

  1 passing (90ms)
```
---

**Reference run 2** ( 932da9d09a018fed4cb830e67090bb994f8539c1 with altered response):

Value of response's nameID is altered to `aaaaaaaa@bbbbb.local`

```patch
diff --git a/test/tests.js b/test/tests.js
index 045db3c..5f7da83 100644
--- a/test/tests.js
+++ b/test/tests.js
@@ -1626,7 +1626,7 @@ describe( 'passport-saml /', function() {
         };
         it( 'valid onelogin xml document should validate', function( done ) {
           var xml = '<samlp:Response xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" ID="R689b0733bccca22a137e3654830312332940b1be" Version="2.0" IssueInstant="2014-05-28T00:16:08Z" Destination="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><samlp:Status><samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/></samlp:Status>' +
-            '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/><ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/><ds:Reference URI="#pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>DCnPTQYBb1hKspbe6fg1U3q8xn4=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>e0+aFomA0+JAY0f9tKqzIuqIVSSw7LiFUsneEDKPBWdiTz1sMdgr/2y1e9+rjaS2mRmCi/vSQLY3zTYz0hp6nJNU19+TWoXo9kHQyWT4KkeQL4Xs/gZ/AoKC20iHVKtpPps0IQ0Ml/qRoouSitt6Sf/WDz2LV/pWcH2hx5tv3xSw36hK2NQc7qw7r1mEXnvcjXReYo8rrVf7XHGGxNoRIEICUIi110uvsWemSXf0Z0dyb0FVYOWuSsQMDlzNpheADBifFO4UTfSEhFZvn8kVCGZUIwrbOhZ2d/+YEtgyuTg+qtslgfy4dwd4TvEcfuRzQTazeefprSFyiQckAXOjcw==</ds:SignatureValue><ds:KeyInfo><ds:X509Data><ds:X509Certificate>'+TEST_CERT+'</ds:X509Certificate></ds:X509Data></ds:KeyInfo></ds:Signature><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">ploer@subspacesw.com</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"><saml:SubjectConfirmationData NotOnOrAfter="2014-05-28T00:19:08Z" Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
+            '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/><ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/><ds:Reference URI="#pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>DCnPTQYBb1hKspbe6fg1U3q8xn4=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>e0+aFomA0+JAY0f9tKqzIuqIVSSw7LiFUsneEDKPBWdiTz1sMdgr/2y1e9+rjaS2mRmCi/vSQLY3zTYz0hp6nJNU19+TWoXo9kHQyWT4KkeQL4Xs/gZ/AoKC20iHVKtpPps0IQ0Ml/qRoouSitt6Sf/WDz2LV/pWcH2hx5tv3xSw36hK2NQc7qw7r1mEXnvcjXReYo8rrVf7XHGGxNoRIEICUIi110uvsWemSXf0Z0dyb0FVYOWuSsQMDlzNpheADBifFO4UTfSEhFZvn8kVCGZUIwrbOhZ2d/+YEtgyuTg+qtslgfy4dwd4TvEcfuRzQTazeefprSFyiQckAXOjcw==</ds:SignatureValue><ds:KeyInfo><ds:X509Data><ds:X509Certificate>'+TEST_CERT+'</ds:X509Certificate></ds:X509Data></ds:KeyInfo></ds:Signature><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">aaaaaaaa@bbbbb.local</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"><saml:SubjectConfirmationData NotOnOrAfter="2014-05-28T00:19:08Z" Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
             '</samlp:Response>';
           var base64xml = Buffer.from( xml ).toString('base64');
           var container = { SAMLResponse: base64xml };
```

```plaintext
passport-saml$ npm run tsc && node_modules/mocha/bin/mocha -g 'valid onelogin xml document should validate'

  passport-saml /
    saml.js / 
      validatePostResponse checks /
        validatePostResponse xml signature checks /
          1) valid onelogin xml document should validate

  0 passing (85ms)
  1 failing

  1) passport-saml /
       saml.js / 
         validatePostResponse checks /
           validatePostResponse xml signature checks /
             valid onelogin xml document should validate:
     AssertionError: expected Error { message: 'Invalid signature' } to not exist
```
Test failed as expected due invalid signature.

---

**Case: `cert` configuration option is not provided at all.**

This could happen
* accidentally or
* due lack of understanding how essential part of SAML's security the signature check is or
* due lack of integration testing and someone pushes temporary debug version into production codebase (or maybe developer is just happy that he/she finally got SAML login to work and does not pay attention to all configuaration parameters, especially to `cert`)

Test case is modified so that SAML is configured without `cert` option and NameID is altered to `aaaaaaaa@bbbbb.local`.

Test should fail due invalid signature or passport-saml should fail fast due missing `cert` configuration option value.

It SHOULD NOT pass so that `passport-saml`'s client application receives altered nameid value...

```patch
diff --git a/test/tests.js b/test/tests.js
index 045db3c..eefaf05 100644
--- a/test/tests.js
+++ b/test/tests.js
@@ -1622,11 +1622,10 @@ describe( 'passport-saml /', function() {
 
         var samlConfig = {
           entryPoint: 'https://app.onelogin.com/trust/saml2/http-post/sso/371755',
-          cert: TEST_CERT,
         };
         it( 'valid onelogin xml document should validate', function( done ) {
           var xml = '<samlp:Response xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" ID="R689b0733bccca22a137e3654830312332940b1be" Version="2.0" IssueInstant="2014-05-28T00:16:08Z" Destination="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><samlp:Status><samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/></samlp:Status>' +
-            '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/><ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/><ds:Reference URI="#pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>DCnPTQYBb1hKspbe6fg1U3q8xn4=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>e0+aFomA0+JAY0f9tKqzIuqIVSSw7LiFUsneEDKPBWdiTz1sMdgr/2y1e9+rjaS2mRmCi/vSQLY3zTYz0hp6nJNU19+TWoXo9kHQyWT4KkeQL4Xs/gZ/AoKC20iHVKtpPps0IQ0Ml/qRoouSitt6Sf/WDz2LV/pWcH2hx5tv3xSw36hK2NQc7qw7r1mEXnvcjXReYo8rrVf7XHGGxNoRIEICUIi110uvsWemSXf0Z0dyb0FVYOWuSsQMDlzNpheADBifFO4UTfSEhFZvn8kVCGZUIwrbOhZ2d/+YEtgyuTg+qtslgfy4dwd4TvEcfuRzQTazeefprSFyiQckAXOjcw==</ds:SignatureValue><ds:KeyInfo><ds:X509Data><ds:X509Certificate>'+TEST_CERT+'</ds:X509Certificate></ds:X509Data></ds:KeyInfo></ds:Signature><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">ploer@subspacesw.com</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"><saml:SubjectConfirmationData NotOnOrAfter="2014-05-28T00:19:08Z" Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
+            '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/><ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/><ds:Reference URI="#pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>DCnPTQYBb1hKspbe6fg1U3q8xn4=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>e0+aFomA0+JAY0f9tKqzIuqIVSSw7LiFUsneEDKPBWdiTz1sMdgr/2y1e9+rjaS2mRmCi/vSQLY3zTYz0hp6nJNU19+TWoXo9kHQyWT4KkeQL4Xs/gZ/AoKC20iHVKtpPps0IQ0Ml/qRoouSitt6Sf/WDz2LV/pWcH2hx5tv3xSw36hK2NQc7qw7r1mEXnvcjXReYo8rrVf7XHGGxNoRIEICUIi110uvsWemSXf0Z0dyb0FVYOWuSsQMDlzNpheADBifFO4UTfSEhFZvn8kVCGZUIwrbOhZ2d/+YEtgyuTg+qtslgfy4dwd4TvEcfuRzQTazeefprSFyiQckAXOjcw==</ds:SignatureValue><ds:KeyInfo><ds:X509Data><ds:X509Certificate>'+TEST_CERT+'</ds:X509Certificate></ds:X509Data></ds:KeyInfo></ds:Signature><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">aaaaaaaa@bbbbb.local</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"><saml:SubjectConfirmationData NotOnOrAfter="2014-05-28T00:19:08Z" Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
             '</samlp:Response>';
           var base64xml = Buffer.from( xml ).toString('base64');
           var container = { SAMLResponse: base64xml };
@@ -1634,7 +1633,7 @@ describe( 'passport-saml /', function() {
           samlObj.validatePostResponse( container, function( err, profile, logout ) {
             try {
               should.not.exist( err );
-              profile.nameID.should.startWith( 'ploer' );
+              profile.nameID.should.startWith( 'aaaaaaaa@bbbbb.local' );
               done();
             } catch (err2) {
               done(err2);
```

```plaintext
passport-saml$ npm run tsc && node_modules/mocha/bin/mocha -g 'valid onelogin xml document should validate'

  passport-saml /
    saml.js / 
      validatePostResponse checks /
        validatePostResponse xml signature checks /
          ✓ valid onelogin xml document should validate

  1 passing (32ms)
```
...but **instead of failing it passed altered nameid to `passport-saml`'s cient application** (even though signature didn't match anymore).

---

**Case: misspelled `cert` configuration option name.**

Basically same as missing `cert` configuration value.

```patch
diff --git a/test/tests.js b/test/tests.js
index 045db3c..03231ff 100644
--- a/test/tests.js
+++ b/test/tests.js
@@ -1622,11 +1622,11 @@ describe( 'passport-saml /', function() {
 
         var samlConfig = {
           entryPoint: 'https://app.onelogin.com/trust/saml2/http-post/sso/371755',
-          cert: TEST_CERT,
+          crt: TEST_CERT,
         };
         it( 'valid onelogin xml document should validate', function( done ) {
           var xml = '<samlp:Response xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" ID="R689b0733bccca22a137e3654830312332940b1be" Version="2.0" IssueInstant="2014-05-28T00:16:08Z" Destination="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><samlp:Status><samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/></samlp:Status>' +
-            '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/><ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/><ds:Reference URI="#pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>DCnPTQYBb1hKspbe6fg1U3q8xn4=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>e0+aFomA0+JAY0f9tKqzIuqIVSSw7LiFUsneEDKPBWdiTz1sMdgr/2y1e9+rjaS2mRmCi/vSQLY3zTYz0hp6nJNU19+TWoXo9kHQyWT4KkeQL4Xs/gZ/AoKC20iHVKtpPps0IQ0Ml/qRoouSitt6Sf/WDz2LV/pWcH2hx5tv3xSw36hK2NQc7qw7r1mEXnvcjXReYo8rrVf7XHGGxNoRIEICUIi110uvsWemSXf0Z0dyb0FVYOWuSsQMDlzNpheADBifFO4UTfSEhFZvn8kVCGZUIwrbOhZ2d/+YEtgyuTg+qtslgfy4dwd4TvEcfuRzQTazeefprSFyiQckAXOjcw==</ds:SignatureValue><ds:KeyInfo><ds:X509Data><ds:X509Certificate>'+TEST_CERT+'</ds:X509Certificate></ds:X509Data></ds:KeyInfo></ds:Signature><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">ploer@subspacesw.com</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"><saml:SubjectConfirmationData NotOnOrAfter="2014-05-28T00:19:08Z" Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
+            '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/><ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/><ds:Reference URI="#pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>DCnPTQYBb1hKspbe6fg1U3q8xn4=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>e0+aFomA0+JAY0f9tKqzIuqIVSSw7LiFUsneEDKPBWdiTz1sMdgr/2y1e9+rjaS2mRmCi/vSQLY3zTYz0hp6nJNU19+TWoXo9kHQyWT4KkeQL4Xs/gZ/AoKC20iHVKtpPps0IQ0Ml/qRoouSitt6Sf/WDz2LV/pWcH2hx5tv3xSw36hK2NQc7qw7r1mEXnvcjXReYo8rrVf7XHGGxNoRIEICUIi110uvsWemSXf0Z0dyb0FVYOWuSsQMDlzNpheADBifFO4UTfSEhFZvn8kVCGZUIwrbOhZ2d/+YEtgyuTg+qtslgfy4dwd4TvEcfuRzQTazeefprSFyiQckAXOjcw==</ds:SignatureValue><ds:KeyInfo><ds:X509Data><ds:X509Certificate>'+TEST_CERT+'</ds:X509Certificate></ds:X509Data></ds:KeyInfo></ds:Signature><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">aaaaaaaa@bbbbb.local</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"><saml:SubjectConfirmationData NotOnOrAfter="2014-05-28T00:19:08Z" Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
             '</samlp:Response>';
           var base64xml = Buffer.from( xml ).toString('base64');
           var container = { SAMLResponse: base64xml };
@@ -1634,7 +1634,7 @@ describe( 'passport-saml /', function() {
           samlObj.validatePostResponse( container, function( err, profile, logout ) {
             try {
               should.not.exist( err );
-              profile.nameID.should.startWith( 'ploer' );
+              profile.nameID.should.startWith( 'aaaaaaaa@bbbbb.local' );
               done();
             } catch (err2) {
               done(err2);
```

```plaintext
passport-saml$ npm run tsc && node_modules/mocha/bin/mocha -g 'valid onelogin xml document should validate'

  passport-saml /
    saml.js / 
      validatePostResponse checks /
        validatePostResponse xml signature checks /
          ✓ valid onelogin xml document should validate

  1 passing (31ms)
```
This **should have failed** due same reasons as "missing `cert` configuration option case".

---

It might be possible to spot incorrectly configured passport-saml stacks by sending SAML login responses with altered content and observing how servers respond to those login responses if services return enought error information to spot such a configuration error. Or in worst case one could manage to get authenticated session with first try. 

`passport-saml` instance without login response signature validation is as good / bad as no authentication at all.

So instead of this:
https://github.com/node-saml/passport-saml/blob/932da9d09a018fed4cb830e67090bb994f8539c1/src/passport-saml/saml.ts#L125-L127

should it be e.g.:
```javascript
    if (!options.cert) {
      throw new Error('Invalid property: cert must not be empty');
    }
```
or
```javascript
    if (!Object.prototype.hasOwnProperty.call(options, 'cert')) {
      throw new Error('Certificate for IdP response validation missing');
    }

    if (!options.cert) {
      throw new Error('Invalid property: cert must not be empty');
    }
```

btw. `passport-saml`'s documentation https://github.com/node-saml/passport-saml/blob/932da9d09a018fed4cb830e67090bb994f8539c1/README.md#security-and-signatures
>It is a good idea to validate the signatures of the incoming SAML Responses. For this, you can provide the Identity Provider's public PEM-encoded X.509 signing certificate using the `cert` configuration key. 

along with how `cert` is not a must have configuration option could give someone an idea that `cert` is not anything special.

cjbarth:
# Description

This addresses the issue that a SAML strategy could be constructed without a certificate, which would allow an insecure situation. The changes needed to force a cert were quite significant. Types had to be adjusted to accommodate these changes. Additionally, changes were made to tests to get them to fail with the correct error. Our tests were in a situation such that once a cert was required the tests would fail as expected, however, the error message that was shown in no way indicated what the reason for the failure actually was. The tests were updated only to get them to actually show the failure saying that "cert is required". This should yield a much better debugging experience going forward.

While the initial version of this PR has many tests failing, I wanted to get some feedback on the approach taken before I fix the tests by adding in a cert. I intend to add in the certs next week to complete this.

This PR also includes the test conditions that were mentioned in #523.

# Checklist:

- Issue Addressed: [Closes #523]
- Link to SAML spec: [N/A]
- Tests included? [Yes]
- Documentation updated? [Yes]


cjbarth:
@markstos, @gugu, @srd90, @HendrikJan: Did you want to have a look at this before I fix all the tests to include a cert?

markstos:
@cjbarth Seems like a good path to me. Thanks for all the work.

s/manditory/mandatory/g

HendrikJan:
@cjbarth your changes look like real improvements, please proceed! I'll gladly update #536 after this has landed.

src/passport-saml/multiSamlStrategy.ts
```diff
@@ -8,35 +8,31 @@ import {
...
-    }
-
-    if (!options.cacheProvider) {
-      options.cacheProvider = new InMemoryCacheProvider({
-        keyExpirationPeriodMs: options.requestIdExpirationPeriodMs,
-      });
-    }
+    const samlConfig: SamlConfig & MultiSamlConfig = {
+      ...options,
+      cert: options.cert ?? "fake cert", // This would never be a valid cert, to it serves to be a good placeholder
```
HendrikJan, line 28: Suggest change "to" to "so" in comment.
// This would never be a valid cert, **so** it serves to be a good placeholder

gugu:
Are you sure change of casing needed? It will start to silently ignore these parameters in existing JS code, which will lead to unexpected behavior changing in module users' code

cjbarth:
It isn't needed, but it does unify our casing. We had a mix of capitalizing acronyms entirely and only capitalizing the first letter. I have followed the .Net convention here (which is very close to TypeScript in most ways): https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN. You might also see this short, reasonable post: https://www.approxion.com/capital-offenses-how-to-handle-abbreviations-in-camelcase/

Thus, in a variable name "SAML" is "Saml" and "RAC" is "Rac". There are already many things in this version that are breaking changes so it seems like a good time to do this. If there are other things of this nature, I don't mind a branch off the 2.x tag and putting in a few more deprecation warnings for things we know are breaking in this code.

src/passport-saml/saml.ts
```diff
@@ -834,7 +847,6 @@ class SAML {
         if (decryptedAssertions.length != 1) throw new Error("Invalid EncryptedAssertion content");
 
         if (
-          this.options.cert &&
           !validSignature &&
           !this.validateSignature(decryptedXml, decryptedAssertions[0], certs!)
```
srd90, line 851: You have changed `certs!` to `certs` i.e. removed `!` character at lines 
https://github.com/node-saml/passport-saml/blob/1c169db06027435167dc0beb4ebde38e2d49195c/src/passport-saml/saml.ts#L799
and
https://github.com/node-saml/passport-saml/blob/1c169db06027435167dc0beb4ebde38e2d49195c/src/passport-saml/saml.ts#L819
so it seems odd that this decrypted assertion signature checking case still has `..., certs!)`.

btw. if you change this certs parameter consider surrounding throw statement with `{  }` (like other similar checks even though that change would be related to code formatting and not to issue that this pull request is addressing (but this pull request's content seems to be already more than just changing cert to truly mandatory configuration option).

srd90:
>Did you want to have a look at this...

@cjbarth Due to my limited typescript/javascript knowledge and quite large changes in this pull request (which include from my point of view ts/js specific "magic") I'm **unable** to see just by browsing the code whether `cert` is now truly mandatory and that there isn't any execution path which would ignore signature checking (also if cert is array of certificates with null/empty values)

src/passport-saml/saml.ts
```diff
@@ -107,100 +107,111 @@ async function promiseWithNameID(nameid: Node): Promise<NameID> {
...
+      additionalAuthorizeParams: this.ctorOptions.additionalAuthorizeParams ?? {},
+      additionalLogoutParams: this.ctorOptions.additionalLogoutParams ?? {},
+      forceAuthn: this.ctorOptions.forceAuthn ?? false,
+      skipRequestCompression: this.ctorOptions.skipRequestCompression ?? false,
+      disableRequestAcsUrl: this.ctorOptions.disableRequestAcsUrl ?? false,
+      acceptedClockSkewMs: this.ctorOptions.acceptedClockSkewMs ?? 0,
+      path: this.ctorOptions.path ?? "/saml/consume",
+      host: this.ctorOptions.host ?? "localhost",
+      issuer: this.ctorOptions.issuer ?? "onelogin_saml",
+      identifierFormat:
```
HendrikJan, line 176: Why not using `??` for `identifierFormat`, while you are using `??` for `authContext`?  
`identifierFormat` and `authnContext` both used `=== undefined` in the old code, so I would expect them to use the same logic in the new code.  
Also in the README the text for both states "if truthy..." which also makes me expect the same kind of logic.

src/passport-saml/saml.ts
```diff
@@ -107,100 +107,111 @@ async function promiseWithNameID(nameid: Node): Promise<NameID> {
...
+        /**
+         * List of possible values:
+         * - exact : Assertion context must exactly match a context in the list
+         * - minimum:  Assertion context must be at least as strong as a context in the list
+         * - maximum:  Assertion context must be no stronger than a context in the list
+         * - better:  Assertion context must be stronger than all contexts in the list
+         */
+        this.ctorOptions.RacComparison = this.ctorOptions.RacComparison ?? "exact";
+        if (
+          ["exact", "minimum", "maximum", "better"].indexOf(this.ctorOptions.RacComparison) === -1
```
HendrikJan, line 205: `["exact", "minimum", "maximum", "better"].includes(this.ctorOptions.RacComparison)` would be more readable in my opinion. But maybe you choose to change as little as possible which would be fine also.

src/passport-saml/saml.ts
```diff
@@ -485,7 +501,7 @@ class SAML {
     Object.keys(additionalParameters).forEach((k) => {
       samlMessage[k] = additionalParameters[k];
     });
-    if (this.options.privateKey) {
+    if (this.options.privateKey != null) {
```
HendrikJan, line 504: I'm not sure what this change does. Does it force you to explicitly set `privateKey` to `null` to not use it?
I would agree, but you'd probably want to add this information to the readme.

HendrikJan:
I'm missing time and experience to fully review these changes.
The changes look well thought through to me and I learn new things while reading it.

cjbarth:
@srd90 I appreciate your feedback. I updated the code in my latest commit to do [better checking](https://github.com/cjbarth/passport-saml/blob/dd72c77720477526a8f9c5f04f6a2749884b110b/src/passport-saml/saml.ts#L693) of the cert property to check for each member of an array. This function is the only one that reads `this.options.cert`. Basically, there is no code path that will allow a cert to go unused. In fact, the SAML class can't even be constructed without [specifying a cert](https://github.com/cjbarth/passport-saml/blob/dd72c77720477526a8f9c5f04f6a2749884b110b/src/passport-saml/saml.ts#L184).

I will add more test to exercise the two new errors that this `certsToCheck` function throws in a future commit.

src/passport-saml/saml.ts
cjbarth: This changes the check from `falsy` to checking for `null` or `undefined`, which both are `falsy`. So, while it does changes things a little, it is now explicitly checking for the absence of a value, not just a value that is `falsy`. If it were `privateKey !== null`, that would be an explicit check for `null`.

src/passport-saml/saml.ts
cjbarth: You are correct. I've reviewed the logic here and found that we can use the same `??` logic. I initially just copied the pre-existing logic. As you can see, there were a lot of places where we played fast-and-lose with what we allowed leaving a lot of room for misconfiguration, even with TypeScript.

cjbarth: After reviewing some tests, it turns out that this logic is important to have written this way. If `null` is specified, then the default is over-ridden, however, if nothing is provided, or `undefined`, then we use the default. Otherwise, we use what is provided. So, for a test to pass which tests to make sure that this block of XML is removed when `null` is passed, this has to remain this way.

cjbarth (commented):
I've reviewed my own code and I think we're good to go on this. Of course, feedback is welcome.

I expect that there is more work that can be done to strengthen the types and to add tests to make sure that we have better coverage, especially of the code that throws exceptions. However, at this point, I believe this will cover `cert` being required. Of course, this is a breaking change.

I will create a follow-up PR to remove support for `privateCert`, since it is deprecated anyway.

src/passport-saml/multiSamlStrategy.ts
```diff
@@ -8,35 +8,31 @@ import {
...
-    }
-
-    if (!options.cacheProvider) {
-      options.cacheProvider = new InMemoryCacheProvider({
-        keyExpirationPeriodMs: options.requestIdExpirationPeriodMs,
-      });
-    }
+    const samlConfig: SamlConfig & MultiSamlConfig = {
+      ...options,
+      cert: options.cert ?? "fake cert", // This would never be a valid cert, so it serves as a good placeholder
```
gugu, line 28: can we add `static requireCertInConstructor = false` in MultiSamlStrategy and `static requireCertInConstructor = true` in SAML.ts ?
and require cert option only if this.constructor.requireCertInConstructor  is true. In this case there is no need for `"fake cert"` string

gugu (changes requested):

src/passport-saml/saml.ts
```diff
@@ -107,100 +107,111 @@ async function promiseWithNameID(nameid: Node): Promise<NameID> {
...
+      get: function () {
+        const cacheProvider = this.options.cacheProvider;
 
-  constructor(options: Partial<SamlOptions>) {
-    this.options = this.initialize(options);
-    this.cacheProvider = this.options.cacheProvider;
+        return cacheProvider;
+      },
+      configurable: true,
+    });
```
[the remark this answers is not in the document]
cjbarth: `configurable: true` allows the property to be deleted and recreated. This allows for lazy initialization of the options. With the change to the constructor I made based on your other feedback, I might be able to simplify this too.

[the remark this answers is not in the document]
cjbarth: Thanks to your suggestion about using a static variable, I've been able to remove all this code and simplify some other code.

cjbarth:
@gugu Are there any more changes you'd like to suggest on this? @HendrikJan or @srd90 Do you have any thoughts on this?

gugu:
No, that's all

src/passport-saml/saml.ts
```diff
@@ -108,99 +108,80 @@ async function promiseWithNameID(nameid: Node): Promise<NameID> {
...
+      RacComparison: (() => {
+        /**
+         * List of possible values:
+         * - exact : Assertion context must exactly match a context in the list
+         * - minimum:  Assertion context must be at least as strong as a context in the list
+         * - maximum:  Assertion context must be no stronger than a context in the list
+         * - better:  Assertion context must be stronger than all contexts in the list
+         */
+        ctorOptions.RacComparison = ctorOptions.RacComparison ?? "exact";
+        if (!["exact", "minimum", "maximum", "better"].includes(ctorOptions.RacComparison)) {
```
forty, line 174: Why do the check in an IIFE ? I would find it clearer to have it after the object initialization

src/passport-saml/saml.ts
```diff
@@ -108,99 +108,80 @@ async function promiseWithNameID(nameid: Node): Promise<NameID> {
...
-     * - better:  Assertion context must be stronger than all contexts in the list
-     */
-    if (
-      !options.RACComparison ||
-      ["exact", "minimum", "maximum", "better"].indexOf(options.RACComparison) === -1
-    ) {
-      options.RACComparison = "exact";
-    }
+    const options = {
+      ...ctorOptions,
```
forty, line 133: I'm not sure I understand everything that's going on.
Why not simply have something like:
```
const DEFAULTS = {
   passive: false,
   disableRequestedAuthnContext: false,
   // all the default values
}

const options = {...DEFAULTS, ...ctorOptions};
```

Note that this can still cause problems with the user explicitly passes an option to null or undefined (like `new SAML({passive: undefined})` ) so some kind of cleanup might be needed beforehand (for JS users - TS users should not have this issue thanks to typing)

Something like
```
// remove all the undefined, and the null for non nullable options.
Object.keys(ctorOptions).forEach((key) => {
   if(ctorOptions[key] === undefined) delete ctorOptions[key];
   if(ctorOptions[key] === null && key !== 'identifierFormat' && key !== 'authnContext') delete ctorOptions[key];
})
```

Enumerating all the properties and their default with `??` as you did works too, but in this case I would not add the `...ctorOptions` at the beginning, which IMO is confusing (and maybe just explicit all properties that don't have any default, if any)

forty (commented):
I made a few generic comments, I don't know the code base too well yet, so it's hard for me to comment on the code organization. 

One thing I noticed it that this PR contains quite a lot of lint fixes (which is great, for example `'!=` => `!==`) and caml case fixes (RAC => Rac) which is also nice, but I would strongly suggest to make a dedicated PR so it can be merged quickly, and then it's easier for everyone to focus on the actual changes (sorry, yet another comment that's a bit generic, that's all I have for now ;) )

src/passport-saml/inmemory-cache-provider.ts
```diff
@@ -37,7 +37,10 @@ export class CacheProvider {
       options.keyExpirationPeriodMs = 28800000; // 8 hours
     }
 
-    this.options = options as CacheProviderOptions;
+    this.options = {
+      ...options,
+      keyExpirationPeriodMs: options.keyExpirationPeriodMs,
```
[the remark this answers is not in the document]
cjbarth: It allows us to preserve all incoming options, while simultaneously converting the option from `partial<CacheProviderOptions>` to `CacheProviderOptions`.

src/passport-saml/types.ts
```diff
@@ -20,40 +20,50 @@ export interface AuthorizeOptions extends AuthenticateOptions {
...
+  attributeConsumingServiceIndex?: string;
   disableRequestedAuthnContext: boolean;
-  authnContext: string | string[];
+  authnContext: string[];
   forceAuthn: boolean;
   skipRequestCompression: boolean;
   authnRequestBinding?: string;
-  RACComparison: RACComparision;
-  providerName: string;
+  RacComparison: RacComparision;
```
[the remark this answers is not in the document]
cjbarth: It should, your right. Done.

forty:
Just a thought: I think that rather than having MutliSaml being a special case of SingleSaml, you should actually have SingleSaml be a special case of MutliSaml (so reverse everything compared to how it's currently, have the main implementation be MultiSaml)

That makes much more sense, since after all SingleSaml is just MutliSaml with n = 1 (so you could have getSamlOptions be a trivial function in this case, that alway return the same config)

src/passport-saml/inmemory-cache-provider.ts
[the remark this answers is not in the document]
forty: I see. I would find it clearer with
```
    this.options = {
      keyExpirationPeriodMs: 28800000; // 8 hours
      ...options
    };
```

src/passport-saml/types.ts
```diff
@@ -63,19 +73,21 @@ export interface SamlOptions extends SamlSigningOptions {
...
-interface StrategyOptions {
+export interface StrategyOptions {
   name?: string;
   passReqToCallback?: boolean;
 }
 
+/**
+ * These options are availble for configuring a SAML strategy
+ */
+export type SamlConfig = Partial<SamlOptions> & MandatorySamlOptions;
```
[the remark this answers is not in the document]
cjbarth: The trick is that `SamlConfig` requires a `cert`, and `MultiSamlConfig` does not. So, we can't pass a `MultiSamlConfig` to the `Strategy` ctor. What is more, `getSamlOptions()` must return an object can can be used in the `Stragegy` ctor, so it must include a `cert`, whereas we don't need a `cert` to construct `MultiSamlStrategy`.

Having said all that, your original comment holds. Done.

src/passport-saml/multiSamlStrategy.ts
```diff
@@ -8,35 +8,35 @@ import {
...
   MultiSamlConfig,
   RequestWithUser,
+  SamlConfig,
   VerifyWithoutRequest,
   VerifyWithRequest,
 } from "./types";
 
 class MultiSamlStrategy extends SamlStrategy {
-  _options: MultiSamlConfig;
+  static newSamlProviderOnConstruct = false;
```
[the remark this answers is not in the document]
cjbarth: And normally I would agree with you, but we can't have an instance field before the call to `super`, and we need this information before doing that. I did however, mark it `readonly`.

See "src/passport-saml/multiSamlStrategy.ts:23:3 - error TS2376: A 'super' call must be the first statement in the constructor when a class contains initialized properties, parameter properties, or private identifiers."

src/passport-saml/saml.ts
cjbarth: The idea was to keep all the logic right together so there wouldn't be two places, even in the same ctor, to look for how a property ends up the way it is. However, I can see how IIFE would actually mutate the incoming obejct and make things slower. Done.

src/passport-saml/saml.ts
cjbarth: Putting in `...ctorOptions` like is done is so that all properties don't have to be explicitly specified, only ones where there are some defaults or special handling. 

Generally, I see what you are after here, but I feel like this logic might be easier to follow than the hoops you describe. It also makes it easier to extend in one place instead of having to go to multiple places. ("I have to remember that if `null` is a valid value, to go to this other block to explicitly allow that.")

cjbarth:
@forty Thank you very much for your review; I've incorporated several changes because of it. Please see my other comments on your review.

As for the MultiSaml and Saml being the way it is, I'm sure you can imagine which one came first and why it is the way it is :) I am not opposed to reversing things, and would welcome a PR to help with that. I think that is out of scope for this PR though. We want to get this landed so we can stop holding up some other PRs.

src/passport-saml/inmemory-cache-provider.ts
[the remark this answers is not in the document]
cjbarth: I've made an adjustment to clarify this code block. Please have a look at this change.

gugu (approved):

cjbarth:
I'm going to land this tomorrow barring any additional feedback so that we can unblock a bunch of other PRs giving others a chance to work with the updated `master` branch over the weekend should they so choose.

src/passport-saml/multiSamlStrategy.ts
```diff
@@ -8,35 +8,35 @@ import {
...
-    if (!options || typeof options.getSamlOptions != "function") {
+    if (!options || typeof options.getSamlOptions !== "function") {
       throw new Error("Please provide a getSamlOptions function");
     }
 
-    if (!options.requestIdExpirationPeriodMs) {
-      options.requestIdExpirationPeriodMs = 28800000; // 8 hours
-    }
-
-    if (!options.cacheProvider) {
```
crossan007, line 29: It seems like this breaks the default `InMemoryCacheProvider` for `MultiSamlStrategy` added via #350  (discussed in #334)

PR #551, opened 2021-03-09
bump xmldom to 0.5.x since all lower versions have security issue

eero3:
# Description

xmldom versions <= 0.4.0 have security issue. Update xmldom dependency to 0.5.x and bump package version. 
More information: https://github.com/xmldom/xmldom/security/advisories/GHSA-h6q6-9hqw-rwfv

 After this fix as been merged, a new version of passport-saml should be released to npm.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "passport-saml",
-  "version": "2.0.5",
+  "version": "2.0.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jokkeliikkanen:
Is there already release scheduled to get this change online?

cjbarth:
The head of `master` currently has semver major changes in it. If we want this in a 2.x release, then we should pull these changes in to a branch off the last 2.x tag. I'd be happy to help with that if needed, though it won't be today :) In any case, we'll end up branching off the 2.x tag to put in some more deprecation notices for the move to 3.x.

cjbarth:
You have to create the PR against the commit you want to fork off of, for example, the latest 2.x tag. You created that PR to merge against `master`.

meetSuraj:
Just wanted to know was this vulnerability exploitable in passport saml? I am going through https://mattermost.com/blog/securing-xml-implementations-across-the-web/. It seems like we need a round trip so that the signature is validated against different input and email is from different. Do we do that in passport-saml? cc @gugu

markstos:
@meetSuraj No one has reported that the vuln was exploitable through passport-saml.  You are welcome to test yourself.

forty:
I can confirm we are parsing data we have serialized ourselves, so it could very well be exploitable. Same for the new xmldom issue. Short term we should update xmldom whenever it's possible. Longer term we should avoid parsing anything we produced. I have plans, but it's a bit of work ;)

PR #552, opened 2021-03-15
bump xmldom to 0.5.x since all lower versions have security issue (#551)

forty:
# Description

This is a cherry-pick of the commit introduced by https://github.com/node-saml/passport-saml/pull/551 on tag v2.0.5, so a v2.0.6 can be released with the security fix (while waiting for a 3.x release).

I think a maintainer will have to create a v2.x branch on the main repo so I can target it with this PR (currently targeting master, which does not make any sense, I know ^^)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "passport-saml",
-  "version": "2.0.5",
+  "version": "2.0.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


cjbarth:
This PR needs to be closed and a new one opened with the requested changes against the latest v2.x tag.

forty:
@cjbarth don't you need to create a branch from tag v2.0.5 first? that's what I'm asking in my PR description, I'm not really sure what a PR against a tag would be otherwise.

cjbarth:
You would make your PR screen look like this, where you've selected a tag, not a branch:
image

forty:
@cjbarth I can watch the diff this way, but not open a PR. That makes sense to me, since I think a branch would be required to point to the result of the merge.

markstos:
@forty I checked-- the "New Pull Request" page does allow selecting a tag as the "base" to submit the PR against.  It looks like the screenshot @cjbarth posted.

forty:
@markstos I can select it indeed, but then the button to open the PR disappears

ezgif-1-6be108cfb511

markstos:
@forty You are right-- it's a bad UI. I've created a "2.x" branch based on the v2.0.5 to submit your PR against. Thank you.

cjbarth:
We actually already had a v2 branch. I was just going to bring it up to date :)

forty:
Ah yes, I did notice that v2 branch, but it did seem very old :) Let me know which one I should use (v2 once @cjbarth has updated it or 2.x)

forty:
@cjbarth I just saw the v2 branch is now updated. I would recommend resetting --hard it to v2.0.5 instead, so that it points to the same commit (or I can use that 2.x branch if that's simpler for everyone)

cjbarth:
I've brought the branch up to date. It now equals the tag v2.0.5. I have a neat little script that is like `git merge -s theirs` to make a merge commit that makes one branch look identical to another and handles all merge conflicts transparently :)

@markstos, do you have an opinion on this. I don't mind resetting `--hard`, but I worry about history. It seems that we should probably just go with the `2.x` branch and if we need even more granularity, we can create a `2.1.x` branch and so on. This way our pattern will be a `v` prefix for tags and no prefix for branches. Thoughts?

forty:
Thanks @cjbarth @markstos and @eero3 to have made it happen.

Note that it seems passport-saml still depends on a vulnerable xmldom version, indirectly via xml-encryption. A patched release has not yet been released to NPM, so we need to wait for xml-encryption upstream to do it. I opened an issue there.

mhassan1:
@forty I've just opened issue https://github.com/node-saml/passport-saml/issues/556 for bumping this package's dependency to `xml-encryption@1.2.2`.

Issue #540, opened 2021-02-16
1.x.x to 2.x.x migration type issue


vbodnia:
Env: Node 14.15.4
Typescript 4.1.3

Currently, I am using passport-saml@1.5.0 with @types/passport-saml@1.1.2. After updating to 2.0.5 I've got a type error in MultiSamlStrategy saying:

> TS2345: Argument of type 'MultiSamlStrategy' is not assignable to parameter of type 'Strategy'. Types of property 'authenticate' are incompatible. Type '(req: RequestWithUser, options: AuthenticateOptions & AuthorizeOptions) => void' is not assignable to type '(this: StrategyCreated<Strategy, Strategy & StrategyCreatedStatic>, req: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, options?: any) => any'. Types of parameters 'req' and 'req' are incompatible. Property 'samlLogoutRequest' is missing in type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>' but required in type 'RequestWithUser'.

The problem is that I'm also extending express requests with custom fields, e.g. user.

cjbarth:
The latest build of `passport-saml` includes its own types and doesn't need anything from DefinitelyTyped.

vbodnia:
@cjbarth In fact, the same issue happens without @types/passport-saml package. I have reinstalled modules with the version 2.0.5 and got the same error.

Mikescops:
I confirm what @vbodnia says, please reopen this issue, it's a current bug in v2.0.5.

cjbarth:
Is this a bug at the `HEAD` of `master`?

vbodnia:
@cjbarth Yes it is:

![image](https://user-images.githubusercontent.com/32565392/113671508-1f1f5880-96bf-11eb-9743-f8c8d4e433af.png)
![image](https://user-images.githubusercontent.com/32565392/113671523-234b7600-96bf-11eb-9dae-a5c5cd738852.png)
![image](https://user-images.githubusercontent.com/32565392/113671538-26466680-96bf-11eb-88cd-9b39c94d03e0.png)

cjbarth:
It seems this might be related to #549. Would you agree?

In any case, I can't see from your screen shot exactly what the problem is. Do you have some sample code I can load up?

fluggo:
The screen shot doesn't show it, but the original description is very similar to mine. I haven't gone back with tweezers yet to figure out where mine went wrong.

[...]

PR #554, opened 2021-03-15
Improve the typing of the Strategy class hierarchy.

forty:
# Description

The goal of this PR is to improve the typing of the Strategy class hierarchy.

The current expected ts error causes problems with some earlier version of TS. 
```
[watch:ts] node_modules/passport-saml/lib/passport-saml/multiSamlStrategy.d.ts(10,5): error TS2416: Property 'generateServiceProviderMetadata' in type 'MultiSamlStrategy' is not assignable to the same property in base type 'Strategy'.
[watch:ts]   Type '(req: Request, decryptionCert: string, signingCert: string, callback: (err: Error, metadata?: string) => void) => void' is not assignable to type '(decryptionCert: string, signingCert?: string) => string'.
```
See https://github.com/node-saml/passport-saml/issues/540 which seems to be affected by this too.

Also, it does not make any sense to have a method override a method with a different signature, just because they have the same name.

### DIFF
--- a/src/passport-saml/index.ts
+++ b/src/passport-saml/index.ts
@@ -1,7 +1,8 @@
 import type { CacheItem, CacheProvider } from "../node-saml/inmemory-cache-provider";
 import { SAML } from "../node-saml";
-import Strategy = require("./strategy");
-import MultiSamlStrategy = require("./multiSamlStrategy");
+import { Strategy, AbstractStrategy } from "./strategy";
+import { MultiSamlStrategy } from "./multiSamlStrategy";
+
 import type {
   Profile,
   SamlConfig,
@@ -12,6 +13,7 @@ import type {
 
 export {
   SAML,
+  AbstractStrategy,
   Strategy,
   MultiSamlStrategy,
   CacheItem,
--- a/src/passport-saml/multiSamlStrategy.ts
+++ b/src/passport-saml/multiSamlStrategy.ts
@@ -1,6 +1,5 @@
-import * as util from "util";
 import { SAML } from "../node-saml";
-import SamlStrategy = require("./strategy");
+import { AbstractStrategy } from "./strategy";
 import type { Request } from "express";
 import {
   AuthenticateOptions,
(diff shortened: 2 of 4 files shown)


cjbarth:
This looks good to me. I have a significant refactor of the type system in #548 which addresses some complex issues surrounding requiring a cert. I'm trying to get that landed before changes like this; though we'll certainly get this landed before we release v3. Feel free to have a look at it and offer feedback.

cjbarth:
Is this something that you wanted fixed in the 2.x branch, or just in `master`?

forty:
@cjbarth I'm actually mostly interested by the 2.x branch for now, since that's what I'm using at the moment. I'll open a PR for that branch too :+1:

cjbarth:
@forty Would you like to resolve the merge conflicts so we can get this landed as part of the upcoming 3.x release?

forty:
Just rebased on master :+1:

test/multiSamlStrategy.spec.ts
```diff
@@ -229,7 +230,7 @@ describe("MultiSamlStrategy#logout", function () {
 describe("MultiSamlStrategy#generateServiceProviderMetadata", function () {
   beforeEach(function () {
     this.superGenerateServiceProviderMetadata = sinon
-      .stub(SamlStrategy.prototype, "generateServiceProviderMetadata")
+      .stub(AbstractStrategy.prototype as any, "_generateServiceProviderMetadata") // force type to any to ignore TS protected property
```
cjbarth, line 233: Why not get rid of `any` and stub `SAML.generateServiceProviderMetadata` instead?

test/multiSamlStrategy.spec.ts
```diff
@@ -20,7 +21,7 @@ describe("MultiSamlStrategy()", function () {
       return { cert: FAKE_CERT };
     }
     const strategy = new MultiSamlStrategy({ getSamlOptions }, noop);
-    strategy.should.be.an.instanceOf(SamlStrategy);
+    strategy.should.be.an.instanceOf(AbstractStrategy);
```
cjbarth, line 24: We should probably include another assert to make sure that it also extends `PassportStrategy`, just to make it clear the chain goes back to something that works with Passport.

cjbarth (changes requested):
As we go along, I'm OK with adding types and making other minor changes in files we're touching to reduce the number of `eslint` errors.

cjbarth:
@forty Would you like to get this rebased and we'll get it landed? If you need help, please let me know.

forty:
@cjbarth I just rebased on latest master :+1:

cjbarth (approved):

cjbarth:
If you want this on 2.x, please create a PR there.

PR #555, opened 2021-03-15
Fix lint npm script to match all files including in src/

forty:
# Description

The current lint rule misses most of the interesting lint errors (in src/) :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 yarn-error.log
 .DS_Store 
 .eslintcache
+.dir-locals.el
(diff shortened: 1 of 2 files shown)


forty:
Note that this increases the number of lint errors from 177 to 252

cjbarth:
Thanks for doing this. BTW, did you test this on Windows? I feel like you might need to replace `'` with `\"` to get this to work on Windows and *nix.

forty:
@cjbarth I don't have a Windows to try that, and I have not idea how quotes work there, so I simply followed your suggestion. It seems consistent with the `lint-watch` script, so that's at least an improvement.

PR #557, opened 2021-03-16
Update xml-crypto to v2.1.1

### LINKED ISSUE #556: [BUG] Dependency on vulnerable version of `xml-encryption`
This package has an exact dependency on `xml-encryption@1.2.1`: https://github.com/node-saml/passport-saml/blob/4d2b90938d6620b823e8f06176a18fb719f3a795/package.json#L56

`xml-encryption@1.2.1` has a dependency on `xmldom@~0.1.15`, which has an XXE vulnerability that has been patched in `xmldom@>=0.5.0`: https://github.com/auth0/node-xml-encryption/blob/9b6df94b0ea30ff7ff836c5e3bf8b328c6a69175/package.json#L25

This package should bump its dependency to `xml-encryption@1.2.3` or higher.

cjbarth:
This was found during an `npm audit`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ node_modules/
 .idea
 yarn-error.log
 .DS_Store 
+.eslintcache
+
--- a/package-lock.json
+++ b/package-lock.json
@@ -757,12 +757,11 @@
       "dev": true
     },
     "axios": {
-      "version": "0.19.2",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-0.19.2.tgz",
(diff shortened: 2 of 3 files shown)


forty:
wonderful! Would you mind updating xml-encryption to ^1.2.3 (latest) while you are at it? They did release the update with the latest xmldom.

Thanks.

forty:
(Else I can create other PRs, as you prefer)

cjbarth:
I think the updates should be different PRs so they appear that way in the change log

PR #562, opened 2021-03-18
Update xml-encryption to v1.2.3

forty:
This update xmldom to a non vulnerable version.


cjbarth:
@forty Did you see that this broke the build?

forty:
@cjbarth Sorry, I have no idea, I cannot reproduce issue here :/ can you?

forty:
@cjbarth hum, running `npm ls` it looks like I messed up the package-lock. I did run an `npm dedup` after my npm install, but I always does, and that usually doesn't cause any trouble :thinking: 

Another weird thing is that it also seems to be messed up on master, but that did no make the build fail :shrug: 

Do you want me to open a PR with a fresh package-lock, regenerated from a clean `npm install` ?

cjbarth:
Let me have a closer look myself. Thanks for the research.

PR #563, opened 2021-03-18
Regenerate a fresh package-lock.

forty:
The current one was messed up by https://github.com/node-saml/passport-saml/pull/560

Sorry for the mess

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,27 +5,27 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.10.4",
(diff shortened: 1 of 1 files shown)


cjbarth:
I have to take a close look at this because "refreshing" the `package-lock.json` will actually change the patch version of our dependencies, and that won't be documented. I think I might roll-back the two PRs that just landed and we can start over.

forty:
Makes sense. And that one seems to break the build too :/ so much for semver reliability :)

At least this time it breaks on my machine too

PR #566, opened 2021-03-18
Update xml-encryption to v1.2.3 (branch 2.x)

forty:
This time no `npm dedup` which seems to mess things up for some reasons.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -760,6 +760,7 @@
       "version": "0.21.1",
       "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.1.tgz",
       "integrity": "sha512-dKQiRHxGD9PPRIUNIWvZhPTPpl1rf/OxTYKsqKUDjBwYylTvV7SjSHJb9ratfyzM6wCdLCOYLzs73qpg5c4iGA==",
+      "dev": true,
       "requires": {
         "follow-redirects": "^1.10.0"
       }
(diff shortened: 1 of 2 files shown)


cjbarth:
So, just to be thorugh, you were on the head of the 2.x branch, then you did a `npm ci`, then you did an `npm install [package]`, and then committed and made this PR. Correct?

forty:
Yes :+1:

forty:
@cjbarth Just a comment on updating minor dependencies: normally update to the package lock is only a "problem" for the build. I believe npm ignore package-lock from packages that you npm install.

It also means that it's a good idea to always do the minor updates, because people will always have the latest version of the dependency that matches the package.json constraints, so it's probably a good idea to run the tests with the latest version.


For reference: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json
```
The difference is that package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the root project.
```

forty:
And I believe most npm libraries (see https://github.com/jaredhanson/passport and https://github.com/expressjs/express which are pupolar and related to the current project) actually don't version their package-lock for that reason.

cjbarth:
You bring up a good point. Let's see if we can adjust some stuff to implement that "best practice". The trick is that we probably want to pin to exact versions for production versions so that we don't have downstream patch releases effect this security library in unexpected ways.

forty:
I think that's a lost battle. Even if you pin exact versions of your direct dependencies, there is no way to pin the transitive dependencies. Even security libraries, such as node-forge which is the crypto lib used by passport-saml (via xml-encryption) put carets everywhere in their package.json https://github.com/digitalbazaar/forge/blob/master/package.json 

I came to think it's for the best. In practice, it's more convenient for the final user of the package to handle the locking, in particular when there are vulnerabilities on some transitive dependencies, and all maintainers are not as reactive to update and/or merge PRs. The only way out of that, is to reduce as much as possible non dev-deps (which is always a good idea anyway) and have only reliable libraries that won't do too big breaking changes in minor updates. For passport-saml, the number of deps is already reasonable, though now that I looked at it I think xml-encryption for example could probably be merged in/vendored (it's only 200 lines of code I think, not worth a dep IMO) and the dependency to passport could be removed for people that don't use express - but sorry I digress ;)

PR #569, opened 2021-03-19
Remove support for deprecated `privateCert`

cjbarth:


gugu (approved):

markstos:
@cjbarth After this change, A `grep` for `privateCert` still returns some hits on the `master` branch in code and types. Is that expected? I thought this PR was removing all traces of `privateCert`

cjbarth:
I don't think that is expected. Perhaps there is something overlooked. I don't see any problems in `master`. Can you point me to them so I can have a further look @markstos ?

PR #571, opened 2021-03-22
Move XML functions to utility module

forty:
# Description

This PR moves most of the XML but non SAML logic to a dedicated file. It tries to improve some types when it's possible. 
The idea is:
1- to isolate dependencies, which I think is always a good idea. Especially when those dependencies are non in TS, this allows to group the "typing danger  zone"
2- shrink saml.ts by moving as much as possible non SAML related logic out.

This PR should not break anything and is done without updating any tests as a result.

Possible follow ups:
- Create a similar crypto-utils for everything crypto (PEM conversion, etc)
- See if xmlbuilder dependency could be dropped in favor of xml2js builder (which is built using xmlbuilder I think). This would avoid depending on 2 different version of xmlbuilder (v15 as a direct dep, v11 via xml2js)

This PR is best reviewed commit per commit.

Let me know what you think.

### DIFF
--- a/src/node-saml/types.ts
+++ b/src/node-saml/types.ts
@@ -3,11 +3,17 @@ import type { CacheProvider } from "./inmemory-cache-provider";
 export type SignatureAlgorithm = "sha1" | "sha256" | "sha512";
 
 export interface SamlSigningOptions {
-  privateKey?: string | Buffer;
+  privateKey: string | Buffer;
   signatureAlgorithm?: SignatureAlgorithm;
   xmlSignatureTransforms?: string[];
   digestAlgorithm?: string;
 }
+
+export const isValidSamlSigningOptions = (
+  options: Partial<SamlSigningOptions>
+): options is SamlSigningOptions => {
+  return options.privateKey != null;
+};
 
 export interface AudienceRestrictionXML {
   Audience?: XMLObject[];
@@ -75,7 +81,7 @@ interface SamlScopingConfig {
  * The options required to use a SAML strategy
  * These may be provided by means of defaults specified in the constructor
  */
-export interface SamlOptions extends SamlSigningOptions, MandatorySamlOptions {
+export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlOptions {
   // Core
   callbackUrl?: string;
   path: string;
--- a/src/node-saml/saml-post-signing.ts
+++ b/src/node-saml/saml-post-signing.ts
@@ -1,42 +1,19 @@
-import { SignedXml } from "xml-crypto";
-import * as algorithms from "./algorithms";
 import { SamlSigningOptions } from "./types";
+import { signXml } from "./xml";
 
 const authnRequestXPath =
   '/*[local-name(.)="AuthnRequest" and namespace-uri(.)="urn:oasis:names:tc:SAML:2.0:protocol"]';
 const issuerXPath =
   '/*[local-name(.)="Issuer" and namespace-uri(.)="urn:oasis:names:tc:SAML:2.0:assertion"]';
-const defaultTransforms = [
-  "http://www.w3.org/2000/09/xmldsig#enveloped-signature",
-  "http://www.w3.org/2001/10/xml-exc-c14n#",
-];
 
 export function signSamlPost(
   samlMessage: string,
   xpath: string,
   options: SamlSigningOptions
 ): string {
-  if (!samlMessage) throw new Error("samlMessage is required");
-  if (!xpath) throw new Error("xpath is required");
-  if (!options) {
-    options = {} as SamlSigningOptions;
-  }
-
-  if (options.privateKey == null) throw new Error("options.privateKey is required");
-
-  const transforms = options.xmlSignatureTransforms || defaultTransforms;
-  const sig = new SignedXml();
-  if (options.signatureAlgorithm) {
-    sig.signatureAlgorithm = algorithms.getSigningAlgorithm(options.signatureAlgorithm);
-  }
-  sig.addReference(xpath, transforms, algorithms.getDigestAlgorithm(options.digestAlgorithm));
-  sig.signingKey = options.privateKey;
-  sig.computeSignature(samlMessage, {
-    location: { reference: xpath + issuerXPath, action: "after" },
-  });
-  return sig.getSignedXml();
+  return signXml(samlMessage, xpath, { reference: xpath + issuerXPath, action: "after" }, options);
 }
 
-export function signAuthnRequestPost(authnRequest: string, options: SamlSigningOptions) {
+export function signAuthnRequestPost(authnRequest: string, options: SamlSigningOptions): string {
   return signSamlPost(authnRequest, authnRequestXPath, options);
 }
--- a/src/node-saml/saml.ts
+++ b/src/node-saml/saml.ts
@@ -3,7 +3,6 @@ const debug = Debug("node-saml");
 import * as zlib from "zlib";
 import * as xml2js from "xml2js";
 import * as crypto from "crypto";
-import * as xmldom from "xmldom";
 import { URL } from "url";
 import * as querystring from "querystring";
 import * as xmlbuilder from "xmlbuilder";
@@ -27,7 +26,7 @@ import {
 } from "./types";
 import { AuthenticateOptions, AuthorizeOptions, Profile, SamlConfig } from "../passport-saml/types";
 import { assertRequired } from "./utility";
-import { decryptXml, validateXmlSignatureForCert, xpath } from "./xml";
+import { decryptXml, parseDomFromString, validateXmlSignatureForCert, xpath } from "./xml";
 
 const inflateRawAsync = util.promisify(zlib.inflateRaw);
 const deflateRawAsync = util.promisify(zlib.deflateRaw);
@@ -715,7 +714,7 @@ class SAML {
     let xml: string, doc: Document, inResponseTo: string | null;
     try {
       xml = Buffer.from(container.SAMLResponse, "base64").toString("utf8");
-      doc = new xmldom.DOMParser({}).parseFromString(xml);
+      doc = parseDomFromString(xml);
 
       if (!Object.prototype.hasOwnProperty.call(doc, "documentElement"))
         throw new Error("SAMLResponse is not valid base64-encoded XML");
@@ -775,7 +774,7 @@ class SAML {
         const encryptedAssertionXml = encryptedAssertions[0].toString();
 
         const decryptedXml = await decryptXml(encryptedAssertionXml, this.options.decryptionPvk);
-        const decryptedDoc = new xmldom.DOMParser().parseFromString(decryptedXml);
+        const decryptedDoc = parseDomFromString(decryptedXml);
         const decryptedAssertions = xpath.selectElements(
           decryptedDoc,
           "/*[local-name()='Assertion']"
@@ -897,7 +896,7 @@ class SAML {
     const data = Buffer.from(container[samlMessageType] as string, "base64");
     const inflated = await inflateRawAsync(data);
 
-    const dom = new xmldom.DOMParser().parseFromString(inflated.toString());
+    const dom = parseDomFromString(inflated.toString());
     const parserConfig = {
       explicitRoot: true,
       explicitCharkey: true,
@@ -1245,7 +1244,7 @@ class SAML {
     container: Record<string, string>
   ): Promise<{ profile?: Profile; loggedOut?: boolean }> {
     const xml = Buffer.from(container.SAMLRequest, "base64").toString("utf8");
-    const dom = new xmldom.DOMParser().parseFromString(xml);
+    const dom = parseDomFromString(xml);
     const parserConfig = {
       explicitRoot: true,
       explicitCharkey: true,
@@ -1293,7 +1292,7 @@ class SAML {
       const encryptedDataXml = encryptedDatas[0].toString();
 
       const decryptedXml = await decryptXml(encryptedDataXml, self.options.decryptionPvk);
-      const decryptedDoc = new xmldom.DOMParser().parseFromString(decryptedXml);
+      const decryptedDoc = parseDomFromString(decryptedXml);
       const decryptedIds = xpath.selectElements(decryptedDoc, "/*[local-name()='NameID']");
       if (decryptedIds.length !== 1) {
         throw new Error("Invalid EncryptedAssertion content");
--- a/src/node-saml/saml.ts
+++ b/src/node-saml/saml.ts
@@ -10,6 +10,7 @@ import * as algorithms from "./algorithms";
 import { signAuthnRequestPost } from "./saml-post-signing";
 import { ParsedQs } from "qs";
 import {
+  isValidSamlSigningOptions,
   AudienceRestrictionXML,
   AuthorizeRequestXML,
   CertCallback,
@@ -355,7 +356,8 @@ class SAML {
     }
 
     let stringRequest = buildXmlBuilderObject(request, false);
-    if (isHttpPostBinding && this.options.privateKey != null) {
+    // TODO: maybe we should always sign here
+    if (isHttpPostBinding && isValidSamlSigningOptions(this.options)) {
       stringRequest = signAuthnRequestPost(stringRequest, this.options);
     }
     return stringRequest;
--- a/src/node-saml/saml.ts
+++ b/src/node-saml/saml.ts
@@ -2,13 +2,11 @@ import Debug from "debug";
 const debug = Debug("node-saml");
 import * as zlib from "zlib";
 import * as xml2js from "xml2js";
-import * as xmlCrypto from "xml-crypto";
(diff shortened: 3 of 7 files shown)


src/passport-saml/xml-utils.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+  fullXml = normalizeNewlines(fullXml);
+  return sig.checkSignature(fullXml);
+};
+
+export const parseDomFromString = (xml: string): Document => {
+  return new xmldom.DOMParser().parseFromString(xml);
+};
+
+export const parseXml2JsFromString = async (xml: string | Buffer): Promise<any> => {
```
forty, line 98: Some improvements can be done on types here I think, but that's going to be for another time.

src/passport-saml/xml-utils.ts
```diff
@@ -0,0 +1,119 @@
...
+
+type SelectedValue = string | number | boolean | Node;
+
+const selectXPath = <T extends SelectedValue>(
+  guard: (values: SelectedValue[]) => values is T[],
+  node: Node,
+  xpath: string
+): T[] => {
+  const result = xmlCrypto.xpath(node, xpath);
+  if (!guard(result)) {
```
forty, line 16: This ensure we get the expected type from the xpath lib at runtime. It should not break anything, and if it does, it's probably for the best.

src/passport-saml/xml-utils.ts
```diff
@@ -38,3 +40,50 @@ export const xpath = {
...
+  util.promisify(xmlenc.decrypt).bind(xmlenc)(xml, { key: decryptionKey });
+
+const normalizeNewlines = (xml: string): string => {
+  // we can use this utility before passing XML to `xml-crypto`
+  // we are considered the XML processor and are responsible for newline normalization
+  // https://github.com/node-saml/passport-saml/issues/431#issuecomment-718132752
+  return xml.replace(/\r\n?/g, "\n");
+};
+
+// This function checks that the |signature| is signed with a given |cert|.
```
cjbarth, line 57: It might be helpful to use the /** */ notation for this comment so that it appears as part of the function help text in IDEs.

cjbarth (changes requested):
Note for future: I'd love to get ride of non-null assertions in favor of an actual runtime validation of the data.

markstos:
I like the idea and recommend dropping "util" from the directory name, leaving it as just "xml". The word "utility" doesn't convey much useful information. It implies minor or miscellaneous functions but I think that's implied by it being an internal library.

cjbarth:
As I look more at what @markstos is saying, it seems we might also be able to consolidate some stuff from `utility.ts` as well since two of the functions there are actually XML functions.

forty:
@cjbarth Yes, true, I did see them then forgot to mention them. I was not 100% sure I should move them, at least one of them (signXmlResponse) seems to be SAML logic to me (it's about a Response element)

I think there is actually a nice factorization opportunity with `saml-post-signing`, so I will move the `signXml` function the xml module and make it generic enough so it can be used by the port signing too.
I might move `signAuthnRequestPost` and `signXmlResponse` back to `saml.ts`, unless you think otherwise?

forty:
(after digging a bit more `signXmlResponse` might actually belong to the test file where it's used)

src/passport-saml/saml-post-signing.ts
```diff
@@ -1,42 +1,19 @@
...
-  }
-
-  if (options.privateKey == null) throw new Error("options.privateKey is required");
-
-  const transforms = options.xmlSignatureTransforms || defaultTransforms;
-  const sig = new SignedXml();
-  if (options.signatureAlgorithm) {
-    sig.signatureAlgorithm = algorithms.getSigningAlgorithm(options.signatureAlgorithm);
-  }
-  sig.addReference(xpath, transforms, algorithms.getDigestAlgorithm(options.digestAlgorithm));
```
forty, line 32: Note that this is different from the reference used bellow for location, but my factorization make it the same (so it becomes `sig.addReference(xpath + issuerXPath, transforms, algorithms.getDigestAlgorithm(options.digestAlgorithm));` )

I don't know if that is good, bad or indifferent, but I notice it doesn't break any tests :) Let me know if I should revert that change.

forty:
Added one commit to move the signXml function to xml module. The change in this commit is slightly more involved than previous commits in this PRs, so review it carefully :) In particular, I probably fixed or broke something, see my comment in the code ;)

I have not touched to the function calling signXml yet, I think their destiny could be decided later, in order not to make that PR bigger that it is already ;)

src/passport-saml/saml-post-signing.ts
cjbarth: I feel like this change could cause a problem, though I would need to do more testing. For example, in the tests "should sign a simple saml request" and "should place the Signature element after the Issuer element", the Issuer is the same, but the Request is different, so I would expect a different signature. After your change I see these two results for the tests:

```xml
<SAMLRequest>
	<saml2:Issuer
		xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" Id="_0">http://example.com
	</saml2:Issuer>
	<Signature
		xmlns="http://www.w3.org/2000/09/xmldsig#">
		<SignedInfo>
			<CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
			<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
			<Reference URI="#_0">
				<Transforms>
					<Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					<Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
				</Transforms>
				<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
				<DigestValue>tmaZyo4Chzy/OjcBg5xxWCGNf1s=</DigestValue>
			</Reference>
		</SignedInfo>
		<SignatureValue>oXiXS9NooDNf7yBRx07P1+aMD2xktkyee/9gzBpdL+MTXsSZ7EBR1OfAM+biVwHaQoC0SW6RmyNYM5CSOAELVGwS9rS1tUB79RqrIgrR0O3GbSX1fsmTULB5Xne2s/8EnY8qPPIGhVcx3zCBW3zNzPwyFePj+U4oMOk52CqSu/mjVuU6pzrbbl4H9jYAg6S23kzJPZGVexfHDkU26wmQPnwisZ1TI9bGwkF2I+wxab2l33LJMnYBBKzPaIM6DxgAFRX+wSnTdGCqxiMcWHD+cr/Mcug8/169Lwaeflt4Bo9+2ojtfZkNtj8ggonR3cAA0KqGCEfAJYUguI5VCtfgyw==</SignatureValue>
	</Signature>
</SAMLRequest>
```

and

```xml
<SAMLRequest>
	<saml2:Issuer
		xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" Id="_0">http://example.com
	</saml2:Issuer>
	<Signature
		xmlns="http://www.w3.org/2000/09/xmldsig#">
		<SignedInfo>
			<CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
			<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
			<Reference URI="#_0">
				<Transforms>
					<Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					<Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
				</Transforms>
				<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
				<DigestValue>tmaZyo4Chzy/OjcBg5xxWCGNf1s=</DigestValue>
			</Reference>
		</SignedInfo>
		<SignatureValue>oXiXS9NooDNf7yBRx07P1+aMD2xktkyee/9gzBpdL+MTXsSZ7EBR1OfAM+biVwHaQoC0SW6RmyNYM5CSOAELVGwS9rS1tUB79RqrIgrR0O3GbSX1fsmTULB5Xne2s/8EnY8qPPIGhVcx3zCBW3zNzPwyFePj+U4oMOk52CqSu/mjVuU6pzrbbl4H9jYAg6S23kzJPZGVexfHDkU26wmQPnwisZ1TI9bGwkF2I+wxab2l33LJMnYBBKzPaIM6DxgAFRX+wSnTdGCqxiMcWHD+cr/Mcug8/169Lwaeflt4Bo9+2ojtfZkNtj8ggonR3cAA0KqGCEfAJYUguI5VCtfgyw==</SignatureValue>
	</Signature>
	<SomeOtherElement/>
</SAMLRequest>
```

You'll note that they are identical except for the `<SomeOtherElement />`. Now, note before this change (literally one I just reverted the last commit you made on your branch):

```xml
<SAMLRequest Id="_0">
	<saml2:Issuer
		xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">http://example.com
	</saml2:Issuer>
	<Signature
		xmlns="http://www.w3.org/2000/09/xmldsig#">
		<SignedInfo>
			<CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
			<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
			<Reference URI="#_0">
				<Transforms>
					<Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					<Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
				</Transforms>
				<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
				<DigestValue>1yis05FW/NgGxi12sn/bW3GP9co=</DigestValue>
			</Reference>
		</SignedInfo>
		<SignatureValue>Oa5ST39rynnUH6XN4tnjoK2luRlKGOq4VHPAKqSgEjzEymTFQRhMwqwQTuFI+AwHSn0qd4wc7GGLIHn0BmUsk/CBZ51nvgjiyTQo+Gkc2/24QlCAwpOM35hgOEaMMvJXgzkFwxvnV/3TGA2J+jrrcQ0q2l6nSuDe27JnCCzbo1vFiHIuWG91pZnS0ZQKnJ593jG5ozo2m2a7l/KvCXIWCGs91KR43IKgmQmOIkVk4i170Ep2trlyj5651LFlT4LShDkkrf4tvWAmeC7rZgf97j58m9vTYXY7zZt5URIvmlE9SZH6NmUdrryZjfZin4Xf7FqpfK/sLzVfBCSLvCse8A==</SignatureValue>
	</Signature>
</SAMLRequest>
```

and

```xml
<SAMLRequest Id="_0">
	<saml2:Issuer
		xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">http://example.com
	</saml2:Issuer>
	<Signature
		xmlns="http://www.w3.org/2000/09/xmldsig#">
		<SignedInfo>
			<CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
			<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
			<Reference URI="#_0">
				<Transforms>
					<Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					<Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
				</Transforms>
				<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
				<DigestValue>5z/cj0WrzI7UGvPSu543FzezwE4=</DigestValue>
			</Reference>
		</SignedInfo>
		<SignatureValue>ggJj62O7r2KwAFyiRtQmWOGu2q1MtLFNcGGCYaeKi5/lmSwhjOqgrLLQlZnQWRIoxhkKw0OLgtT9dIeZC3+TMZtolLQ9OM0pQARz8svJuYQUd4ti71hoIRTRzgzEXbOvpyDoqXaJMeZeveidAg/DHIIATpCUwqy1soUPxiHXdweXJ8BYrNoWjFBKLULbBNmTlYEdeQqYOE3TcCuvCDYOdbQlTKRPJlC5eVz+SIRb8q6c9sHTzmFuiquFezQsroY3MEvHjO5jPQ1L3L1drVwWBeF8TNoNsHcJ8aMMMJqHzjiSDeAJZnq6G3VOisZrJVcSJWipAd383EyuhDy3Zo5PJw==</SignatureValue>
	</Signature>
	<SomeOtherElement/>
</SAMLRequest>
```

Here you'll note that the signature is different. So, this change that you made results in a signature that doesn't actually sign the contents of the Response, but just the Issuer, which means that any change we want can be made to the response and the signature would still check.

So, we need to do two things:

1. Put in a new test to catch this case
2. Make sure that the code changes that your making don't cause this problem

cjbarth (changes requested):
I really like where you are going with this, but your last commit on this branch is a step too far and breaks stuff. I'd like to get a test written that catches this. It should be an easy test. If we take `"should sign a simple saml request"` and `"should place the Signature element after the Issuer element"` and compare the signatures generated with the same settings, they shouldn't be equal.

src/passport-saml/saml-post-signing.ts
forty: I had figured out that something like this could be happening, that makes sense now. I'll add the test case and fix it 👍

forty: should be better now

forty:
I improved the existing tests, I don't see any reason to resort to regexp parsing xml if everything is deterministic and we can check the full result. I confirmed the tests failed with the previous version and fixed.

test/saml-post-signing-tests.spec.ts
```diff
@@ -1,27 +1,129 @@
...
 describe("SAML POST Signing", function () {
-  it("should sign a simple saml request", function () {
+  it("should sign a simple saml request", async function () {
     const xml =
       '<SAMLRequest><saml2:Issuer xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">http://example.com</saml2:Issuer></SAMLRequest>';
     const result = signSamlPost(xml, "/SAMLRequest", { privateKey: signingKey });
-    result.should.match(/<DigestValue>[A-Za-z0-9/+=]+<\/DigestValue>/);
-    result.should.match(/<SignatureValue>[A-Za-z0-9/+=]+<\/SignatureValue>/);
+    const doc = await parseXml2JsFromString(result);
+    doc.should.be.deepEqual({
```
cjbarth, line 14: This is certainly more robust than what we had before. I think this is why a lot of tests compare to a static file: this makes the individual test not fit on a single screen. I can't think of a great solution right now, so I'm not looking for a change, but am open to ideas, perhaps in a follow-on PR.

cjbarth (approved):
This looks good to me. I'll leave this open for a day or so before landing it to let others review.

cjbarth:
@markstos This looks good to me, but I wanted to leave it open to give you a chance to look in case you're interested.

cjbarth:
@forty If you're OK with it, I'd like to hold off on landing this until #574 lands. We have been trying to get #574 done for a while and I'd like to not complicate the already herculean efforts to split out `saml.ts` to its own project with conflicts from this landing. That will likely mean that you're work will need to have some merge conflicts resolved.

I still want to work with you and #574 to make sure these both land before we release 3.x in a few weeks if you're up for it.

forty:
Haha, well I cannot say I'm overly enthusiastic to having to tackle that conflict that promises to be huge, but I'll trust you in choosing the order of merge and will do my best to rebase quickly if needed :)

forty:
(and I have to add, I'm one of those that would be more then happy to get rid of the passport dependencies, so i know it's for a good cause :) )

cjbarth:
@forty We've landed that big PR to start the separation of node-saml from the passport strategy parts. If you would like to revisit your changes here, I'll try to keep up with you in reviewing them.

forty:
@cjbarth Done for that one too! It was easier than expected :sweat_smile:

cjbarth (commented):
It looks good as it is, but I notice that there may be a duplicate helper function. See below.

src/node-saml/types.ts
```diff
@@ -3,12 +3,18 @@ import type { CacheProvider } from "./inmemory-cache-provider";
...
 
 export interface SamlSigningOptions {
-  privateKey?: string | Buffer;
+  privateKey: string | Buffer;
   signatureAlgorithm?: SignatureAlgorithm;
   xmlSignatureTransforms?: string[];
   digestAlgorithm?: string;
 }
 
+export const isValidSamlSigningOptions = (
```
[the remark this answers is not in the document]
forty: It might work too though that one is more specific. I have to admit, I like type guards and probably over use them, so I'm happy to try switching it to assertRequired if you feel that's better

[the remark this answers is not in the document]
forty: After having a closer look, it looks like assertRequired throws, and I don't think that's what we want here

cjbarth (approved):

Issue #575, opened 2021-03-25
[BUG] XML-encoded carriage returns are not resolved correctly


mhassan1:
**Background**

Earlier this week, Google rolled out a change to their GSuite SAML Apps IdP where the SAMLResponse contains XML-encoded carriage returns inside the `SignatureValue` and `X509Certificate` blocks.

For example, they started sending this (notice the `&#13;` at the end of each line):
```xml
<ds:SignatureValue>nLcgyGLUW61YKksmQ1zc/WlV4/iOVlNWZrT13wUWxMwarc38L0Z1b72gTZjfYwH6t06qm4bNvTKc&#13;
B27ctMgEFvnEBwRuYUluLdhUm1HRPVtKflypI3HDqNtv4wYhbxhL11Yqa/Qhg/cV/VvYA1zMpNax&#13;
SEk10RTllBLmKrQbkmED9vnS1184IkynvpA8rJNqvkwkjVAhQKFIOLr0GVuhgIKP6obTbgtJmXlQ&#13;
Ie2oXsrIegeQqSTk64qsDG5MwG/wm5/26+IBC9Dn8aobLQ798xMZmoXNMyMMfUDK6cbMb2LK3hDR&#13;
NiiwuVSaBboGWTovciv+ui+odUbKHpLhVzJT4g==</ds:SignatureValue>
```

At first, I thought this must be a bug introduced on their end, but I saw that both [samltool.com](https://www.samltool.com/validate_response.php) and [SimpleSAMLphp](https://simplesamlphp.org/) considered that to be a valid SAMLResponse. I also noticed that other Java-based (Apache Santuario) IdPs will provide a SAMLResponse like that when a certain system property is unset ([link](https://issues.apache.org/jira/browse/SANTUARIO-494)). I could not find anywhere in the SAML or XML spec that mentions this, but I did notice that `DOMParser` correctly replaces those `&#13;` with actual carriage returns. We can confirm it like this:

```js
const { DOMParser } = require('xmldom')
new DOMParser()
  .parseFromString('<abc>123&#13;456</abc>')
  .toString()
  .replace(/\r/g, '\\r') // just for display purposes here
// -> `123\r456`
```

I believe the root of the problem is that `validateSignatureForCert` passes raw XML (that has never been passed through a `DOMParser`) to `sig.checkSignature(fullXml)`.

**To Reproduce**

[...]

PR #576, opened 2021-03-25
Resolve XML-encoded carriage returns during signature validation

mhassan1:
# Description

See https://github.com/node-saml/passport-saml/issues/575 for background. This PR duplicates https://github.com/node-saml/passport-saml/pull/578 for `master`.

This adds XML normalization to `validateSignatureForCert` before sending it to `xml-crypto`. It uses `DOMParser` for normalization tasks such as replacement of XML-encoded entities (`&#13;`) with their actual representations (carriage return).

# Checklist:

- Issue Addressed: [X]
- Link to SAML spec: [ ]
- Tests included? [X]
- Documentation updated? N/A

PR #578, opened 2021-03-29
Resolve XML-encoded carriage returns during signature validation (2.x)

mhassan1:
# Description

This is a 2.x backport of https://github.com/node-saml/passport-saml/pull/576.

Resolves https://github.com/node-saml/passport-saml/issues/575. See that issue for background.

This adds XML normalization to validateSignatureForCert before sending it to xml-crypto. It uses DOMParser for normalization tasks such as replacement of XML-encoded entities (&#13;) with their actual representations (carriage return).

# Checklist:

- Issue Addressed: [X]
- Link to SAML spec: [ ]
- Tests included? [X]
- Documentation updated? N/A

### DIFF
--- a/src/passport-saml/saml.ts
+++ b/src/passport-saml/saml.ts
@@ -730,6 +730,8 @@ class SAML {
     if (totalReferencedNodes.length > 1) {
       return false;
     }
+    // normalize XML to replace XML-encoded carriage returns with actual carriage returns
+    fullXml = this.normalizeXml(fullXml);
     fullXml = this.normalizeNewlines(fullXml);
     return sig.checkSignature(fullXml);
   }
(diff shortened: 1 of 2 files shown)


src/passport-saml/saml.ts
```diff
@@ -730,6 +730,8 @@ class SAML {
     if (totalReferencedNodes.length > 1) {
       return false;
     }
+    // normalize XML to replace XML-encoded carriage returns with actual carriage returns
+    fullXml = this.normalizeXml(fullXml);
     fullXml = this.normalizeNewlines(fullXml);
```
cjbarth, line 735: Now that we are using `xmldom` to normalize, do we need to call `normalizeNewlines()`?

mhassan1: Yes, we still need it. If we remove `normalizeNewlines()`, the tests will fail. `xmldom` will normalize the encoded carriage returns into `\r`, but we (as the XML processor) are still responsible for normalizing the `\r` into `\n`.

cjbarth (approved):

mhassan1:
Is there a planned 2.x release that will include this?

cjbarth:
Done @mhassan1

Issue #429, opened 2020-03-26
better integration with xml-encryption 1.1.0 release


markstos:
I'd like to address multiple issues with our relationship with the xml-encryption package:

 1. xml-encryption 1.0 disabled some insecure algorithms. By extension we did too. But they expose an option to re-enable the insecure algorithms if you need them. Maybe we should to.
 2. xml-encryption 1.1.0 offers an option to warn if there are insecure algorithms. By combining this option with the one above, a project could start to issue warnings for insecure algorithms without failing, easing the upgrade. We could expose both options. 
 3. We have a place where list a number of algorithms which we claim should be in sync with xml-encryption, but we are not: https://github.com/bergie/passport-saml/blob/fb1bda0d19e5b206e772562bf0ab6d1a1ce96ae4/lib/passport-saml/saml.js#L1264 xml-encryption supports two "GCML" algorithms which we don't, but should.  Also, we should not include "tripledes-cbc" as a supported algorithm if insecure algorithms are disabled. Currently they are, but point 1. above suggests we should.

cjbarth:
@markstos Is this something that we wanted to address for the upcoming v3 release?

markstos:
@cjbarth We could enable the first phase-- issuing warnings by default for algorithms that are insecure that we will be dropping in a future release. 

Also, if we are able to enable the GCML algorithms to be in sync with the spec, we should (assuming they are still considered secure). 

In version three, we could have an "insecure algorithms option" that defaults to true-with-warning, with the option to go ahead and disable the insecure algorithms. 

So yes, I think it would good to address aspects of this in the next version.

cjbarth:
I was thinking that we could do like we did with a few other things and release a small PR for 2.x that has warnings and then just use the switch for the 3.x branch that we're building now.

markstos:
That works too.

cjbarth:
I'll gladly review a PR from you on that and I can probable make the 2.x deprecation version that matches the changes you make.

PR #584, opened 2021-04-28
Enforce more secure XML encryption

cjbarth:
Add newer encryption algorithms and remove insecure ones.

# Checklist:

- Issue Addressed: [ Addresses #429 ]
- Link to SAML spec: [ n/a ]
- Tests included? [ yes ]
- Documentation updated? [ n/a ]


markstos:
Here's a reference to Triple DES being considered insecure now: 
https://security.stackexchange.com/questions/146710/is-the-3des-algorithm-secure

Here's a reference about AES 256 GCM being considered secure:
https://security.stackexchange.com/questions/184305/why-would-i-ever-use-aes-256-cbc-if-aes-256-gcm-is-more-secure

And regarding the final addition, AES-128-GCM, this reference says it's preferable in some cases over AES-256-GCM for performance reasons:

https://crypto.stackexchange.com/questions/77750/why-gcm-operation-mode-with-aes-128-is-recomended-and-can-we-use-aes-192-and-aes

Given the above, I approve this for 3.0.0, which is making other breaking changes anyway.

PR #589, opened 2021-05-11
Add Node 16 support; drop Node 10

cjbarth:


markstos:
Confirmed that Node.js 10 is EOL'ed: https://endoflife.date/nodejs

PR #590, opened 2021-05-11
Update all dependencies to latest

cjbarth:
All dependencies have been updated to their latest version, including semver-major releases. All tests continue to pass and all changelogs for dependencies were reviewed and none of the breaking changes were impactful. The bulk of the reason for the major version bumps in dependencies was due to change in supported Node versions.


markstos:
Running the test suite and reviewing changelogs is the same QA I would have done myself, so I give this a pass.

PR #591, opened 2021-05-13
Remove deprecated field `privateCert` from README, tests

sdaikichi:
`privateCert` looks like it has been removed.
I hope this PR is accepted.


cjbarth:
Good catch @sdaikichi. However, further searching indicates that there are several more places where this is wrong. Would you be willing to update your PR with the occurrences in both `README.md` files and `tests.spec.ts`?

cjbarth:
@sdaikichi You'll probably need to merge the HEAD of `master` in to your branch to get it to pass tests.

cjbarth (approved):

Issue #543, opened 2021-02-19
[BUG] Possibility to overwrite profile fields with data from attributes


gtokarz:
Currently attributes are mapped to profile without any check, which might at best cause minor issues with previously present properties from profile object.

**To Reproduce**

Configure Idp/craft a SAML response with attribute Name set to i.e issuer or other profile field set before attribute mapping. Mapped profile object should now have overwritten issuer field.

**Expected behavior**
It should check if property already exists on the profile object and only map values when key is not present.
https://github.com/node-saml/passport-saml/blob/f1a436fa243080f8cfcf06fd0f40ce3af396c8b6/src/passport-saml/saml.ts#L1319

**Environment**

- Node.js version: 12.16.1
- passport-saml version: 2.0.5

cjbarth:
@gtokarz would you be willing to create a PR with tests for this?

gtokarz:
@cjbarth Yes, I might have some time on Monday to do so

cjbarth:
@gtokarz We are preparing a 3.0 release. Would you be interested in getting this change up for PR in the next few weeks to make it in to 3.x?

kriss1897:
@cjbarth @gtokarz Is it okay if I raise a PR for this?

cjbarth:
@kriss1897 sure, anything with a `pr-welcome` tag we are happy to have someone help with. Please make sure to include tests for this.

gtokarz:
yes, go ahead with PR

kriss1897:
@cjbarth @gtokarz I've raised a PR for fixing this, however I'm unable to assign any reviewers. Please take a took when you guys have time.

PR #593, opened 2021-05-15
Fix: Conflicting profile properties between profile and attributes

kriss1897:
# Description
https://github.com/node-saml/passport-saml/blob/f1a436fa243080f8cfcf06fd0f40ce3af396c8b6/src/passport-saml/saml.ts#L1319
It is currently possible to override some default properties of the `profile` object in `processValidlySignedAssertionAsync` method.
This can be done if any property like `Issuer` is also present in the attribute mapping. The expected behavior here should be that the properties (`Issuer`, `inResponseTo`, `sessionIndex`, `nameID`, `nameIDFormat`, `nameQualifier`, `spNameQualifier`) will probably already be present in the profile, and in that case they shouldn't be overridden by the same named attribute in attribute mapping.

This issue can be handled gracefully by simply skipping the value from attribute mapping or an error can be thrown. This PR handles the issue gracefully. 

# Checklist:

- Issue Addressed: [#543]
- Link to SAML spec: [ ]
- Tests included? Yes
- Documentation updated? Not Required

### DIFF
--- a/src/node-saml/saml.ts
+++ b/src/node-saml/saml.ts
@@ -1170,13 +1170,21 @@ class SAML {
       };
 
       if (attributes) {
+        const profileAttributes: Record<string, unknown> = {};
+
         attributes.forEach((attribute) => {
           if (!Object.prototype.hasOwnProperty.call(attribute, "AttributeValue")) {
             // if attributes has no AttributeValue child, continue
             return;
           }
+
           const name = attribute.$.Name;
-          const value = attribute.AttributeValue;
+          const value =
+            attribute.AttributeValue.length === 1
+              ? attrValueMapper(attribute.AttributeValue[0])
+              : attribute.AttributeValue.map(attrValueMapper);
+
+          profileAttributes[name] = value;
 
           // If any property is already present in profile and is also present
           // in attributes, then skip the one from attributes. Handle this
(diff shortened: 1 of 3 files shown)


src/node-saml/saml.ts
```diff
@@ -1175,11 +1175,20 @@ class SAML {
             // if attributes has no AttributeValue child, continue
             return;
           }
+          const name = attribute.$.Name;
           const value = attribute.AttributeValue;
+
+          // If any property is already present in profile and is also present
+          // in attributes, then skip the one from attributes. Handle this
+          // conflict gracefully without returning any error
+          if (Object.prototype.hasOwnProperty.call(profile, name)) {
```
cjbarth, line 1184: Do you think there is value is putting all these values in a child object on the profile in addition to cloning them directly to the profile object where there is no conflict? This way, nothing is lost, but we also don't break existing behavior.

kriss1897: Yes. I think putting all attributes in separate child property would be useful. And only clone the properties that do not exist on profile.

kriss1897: Update PR with the requested changes

cjbarth:
This looks good to me. However, either we need to make this PR only against the [node-saml](https://github.com/node-saml/node-saml) project, or we need to land this PR in both this and that project.

cjbarth:
Sounds good to me. As soon as I see a corresponding PR in the [node-saml](https://github.com/node-saml/node-saml) project, we'll land it here.

kriss1897:
@cjbarth Added same PR on node-saml repo. https://github.com/node-saml/node-saml/pull/5

cjbarth (approved):

Issue #602, opened 2021-05-27
Feature: add facility in config to add `<Extensions>` element in SAML request


kdhttps:
I've checked all the [config parameters](https://github.com/node-saml/passport-saml#config-parameter-details) but didn't find any config to add `<Extensions>` element in SAML request like below:

```xml
<samlp:AuthnRequest AssertionConsumerServiceURL="https://iam.test.fi/oxauth/postlogin" Destination="https://testidp.fi/idp/profile/SAML2/Redirect/SSO" ID="_3a5525fc-f5cc-46af-add6-70bbc27ecebf" IssueInstant="2021-05-19T11:05:04Z" Version="2.0" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol">
<saml:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">https://iam.test.fi/oxauth/1</saml:Issuer>

<samlp:Extensions>
<vetuma xmlns="urn:vetuma:SAML:2.0:extensions"><LG>fi</LG></vetuma>
</samlp:Extensions>

<samlp:NameIDPolicy AllowCreate="true"/>
<samlp:RequestedAuthnContext Comparison="exact">
<saml:AuthnContextClassRef xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">http://testidp.fi/2017/loa3</saml:AuthnContextClassRef>
</samlp:AuthnRequest>
```

Thank you!

srd90:
There is unfinished Extensions related pull request #147

---

Not directly related to your issue but more like a FYI:

Based on your example (vetuma language code extension) you are most probably going to use suomifi IdP and that you are using HTTP-Redirect binding for authn request.

According to suomifi IdP's `Technical interface description` https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70 (referenced May 28 2021) you could send language code also as query parameter:
>The language code can also be passed as the variable's locale value in the HTTP protocol (similar to the event code passed in the RelayState variable).
>
>The transmission of the language code thus only works in the HTTP-Redirect profile in such a way that it is the last query parameter.
>
>This method should only be used if, for technical reasons, the language code cannot be included in the SAML message.

Be aware that the same technical interface description document also says about SLO that:
>The e-service must be able to both initiate the log-out process and handle the log-out request that comes from the Identification.

[...]

PR #607, opened 2021-06-08
Feature: add facility in config to add `<Extensions>` element in SAML request

kdhttps:
## Issue
close https://github.com/node-saml/passport-saml/issues/602

## Description

Add support in `passport-saml` where it will allow admin to configure fully customise `<Extensions>` element and add it into SAML Request.

`<Extensions>` element are use to request custom attributes and setting. Its structure and values different as as IDP configurations and requests.

## Use cases
1.  Use `<Extensions>` element to request attributes/data. [More details, oasis-open.org SAML V2.0 Protocol Extension for Requesting Attributes per Request](http://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html)
2. Use `<Extensions>` element to request with some custom config. For Example: In [suomi.fi](https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70) idp, its allows SP to request with custom language code by passing it `<Extensions>` element. [More details](https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70)

## Changes

- `src/node-saml/types.ts` : added `extensions` optional variable. The type is `any` and it accept [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type values because as per use cases and IDPs', there are many cases and each use cases have different request format/data. Please let me know what is your view on this.
-  `src/node-saml/saml.ts`: added it in `request`. As it has already XMLBuilder values so no need to do any processing.
-  Test cases added
-  Docs updated 
-  No Breaking Changes

# Checklist:

- [x] Issue Addressed
- [x] Link to SAML spec
- [x] Tests included
- [x] Documentation updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,7 @@ type Profile = {
 - `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
 - `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
 
+- **Extensions**
 - `extensions`: Optional, The extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://www.oasis-open.org/committees/download.php/55790/Connectis%20_protocol_extension_draft.pdf). It accept fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type. There are many possible value for `extensions` element. You can use [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) to request attributes/data as per your need in SAML Request.
 
 ```javascript
--- a/README.md
+++ b/README.md
@@ -12,8 +12,6 @@ Passport-SAML has been tested to work with Onelogin, Okta, Shibboleth, [SimpleSA
 
     $ npm install passport-saml
 
-/
-
 ## Usage
 
 The examples utilize the [Feide OpenIdp identity provider](https://openidp.feide.no/). You need an account there to log in with this. You also need to [register your site](https://openidp.feide.no/simplesaml/module.php/metaedit/index.php) as a service provider.
@@ -175,6 +173,24 @@ type Profile = {
 - `logoutUrl`: base address to call with logout requests (default: `entryPoint`)
 - `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
 - `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
+
+- `extensions`: Optional, The extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://www.oasis-open.org/committees/download.php/55790/Connectis%20_protocol_extension_draft.pdf). It accept fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type. There are many possible value for `extensions` element. You can use [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) to request attributes/data as per your need in SAML Request.
+
+```javascript
+// Example
+extensions: {
+  "md:RequestedAttribute": {
+    "@isRequired": "true",
+    "@Name": "Lastname",
+  },
+  vetuma: {
+    "@xmlns": "urn:vetuma:SAML:2.0:extensions",
+    LG: {
+      "#text": "sv",
+    },
+  },
+},
+```
 
 ### Provide the authentication callback
 
--- a/src/node-saml/types.ts
+++ b/src/node-saml/types.ts
@@ -123,4 +123,5 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
 
   // extras
   disableRequestAcsUrl: boolean;
+  extensions?: any;
 }
--- a/test/passport-saml/capturedSamlRequests.spec.ts
+++ b/test/passport-saml/capturedSamlRequests.spec.ts
@@ -859,6 +859,73 @@ const capturedSamlRequestChecks: SamlCheck[] = [
       },
     },
   },
+  {
+    name: "Config with Extensions",
+    config: {
+      cert: FAKE_CERT,
+      extensions: {
+        "md:RequestedAttribute": {
+          "@isRequired": "true",
+          "@Name": "Lastname",
+        },
+        vetuma: {
+          "@xmlns": "urn:vetuma:SAML:2.0:extensions",
+          LG: {
+            "#text": "sv",
+          },
+        },
+      },
+    },
+    result: {
+      "samlp:AuthnRequest": {
+        $: {
+          "xmlns:samlp": "urn:oasis:names:tc:SAML:2.0:protocol",
+          Version: "2.0",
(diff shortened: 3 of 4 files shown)


cjbarth:
Once this is approved, we'll need it duplicated over on the [node-saml](https://github.com/node-saml/node-saml) project before it can be landed so that we keep these in sync.

README.md
```diff
@@ -176,6 +174,25 @@ type Profile = {
 - `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
 - `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
 
+- **Extensions**
+- `extensions`: Optional, The extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://www.oasis-open.org/committees/download.php/55790/Connectis%20_protocol_extension_draft.pdf). It accept fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type. There are many possible value for `extensions` element. You can use [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) to request attributes/data as per your need in SAML Request.
```
cjbarth, line 178: The grammar here is a little off, but it is still understandable. We can fix it all at once later.

README.md
```diff
@@ -176,6 +174,25 @@ type Profile = {
...
+
+```javascript
+// Example
+extensions: {
```
cjbarth, line 182: I feel like the name `extensions` is too generic and might logically collide with something else. The documentation referenced continuously refers to "SAML Protocol Extension for Requesting Attributes per Request", so clearly they are leaving room for some other type of extension to do some other thing.

Would it make sense to call the section `samlExtensions` to make it more clear that these are extension that go directly into the SAML document and not some extension to `node-saml` or `passport-saml` functionality?

README.md
cjbarth: We might want to change the reference to the approved spec, or at least add it. The PDF is a good reference, but is listed as a draft. Here is the official spec that we want to support: https://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html

README.md
```diff
@@ -176,6 +174,25 @@ type Profile = {
...
+  "md:RequestedAttribute": {
```
cjbarth, line 183: Expressing XML as JavaScript is prone to mistakes, and it can be hard to debug. Would there be another way to express this? Perhaps we should require the actual XML document fragment, or XMLBuilder object. This would force problems to be sorted out before they reach `passport-saml`.

src/node-saml/types.ts
```diff
@@ -123,4 +123,5 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
 
   // extras
   disableRequestAcsUrl: boolean;
+  extensions?: any;
```
cjbarth, line 126: This doesn't help the consumer very much. If there is a way that we could specify something more helpful, that would be good. Even if we just know it will be a string, or an object, that would help. My previous comments about this being XML would be even better if it can be done.

test/passport-saml/capturedSamlRequests.spec.ts
```diff
@@ -859,6 +859,73 @@ const capturedSamlRequestChecks: SamlCheck[] = [
       },
     },
   },
+  {
```
cjbarth, line 862: Nice clean addition of another test leveraging the framework. Thanks!

cjbarth (changes requested):
I think the biggest discussion point here is that we are taking pretty much anything and shoving it in the SAML request after trying to convert it and then hoping it all turns out OK. I think that is prone to error and open to potential misuse (though that is much more unlikely for sure).

README.md
kdhttps: @cjbarth first of all thank you for review and sorry about my English. I agreed to add official docs link instead of PDF. I'll update it.

README.md
kdhttps: agreed @cjbarth , `samlExtensions` is more clear. i'll make a change.

test/passport-saml/capturedSamlRequests.spec.ts
kdhttps: Thank you :)

src/node-saml/types.ts
kdhttps: I tried `object` type but linting not allow. Need to think on it. Give me some time.

README.md
kdhttps: Need to think on it. Give me some time.

kdhttps:
> I think the biggest discussion point here is that we are taking pretty much anything and shoving it in the SAML request after trying to convert it and then hoping it all turns out OK. I think that is prone to error and open to potential misuse (though that is much more unlikely for sure).

We may build some custom types. For now, I am seeing two possibilities there, one is on official docs and second on `suomi` IDP requirements. Need to think on this.

README.md
cjbarth: Don't worry about your level of English. Learning a second language is not easy. I'll fix it up for you as a last item before we land this.

cjbarth:
I don't know that custom types are the way to go, as there is probably too much variability, but enforcing XML is probably a good idea. I'm eager to hear your ideas for improvement.

kdhttps:
> I don't know that custom types are the way to go, as there is probably too much variability, but enforcing XML is probably a good idea. I'm eager to hear your ideas for improvement.

Agreed there is too much variability. XML is good idea but could you please describe, how exactly? I mean you are thinking to set type string to `samlExtensions` and pass XML string like below:

```js
{
samlExtensions: ` <samlp:Extensions>
<req-attr:RequestedAttributes>
    <md:RequestedAttribute isRequired="true" Name="urn:oid:2.5.4.4"
      NameFormat=”urn:oasis:names:tc:SAML:2.0:attrname-format:uri” />
    <md:RequestedAttribute isRequired="true" Name="urn:oid:2.5.4.42"
      NameFormat=”urn:oasis:names:tc:SAML:2.0:attrname-format:uri” />
    <md:RequestedAttribute Name="urn:oid:0.9.2342.19200300.100.1.3"
      NameFormat=”urn:oasis:names:tc:SAML:2.0:attrname-format:uri” />
    <md:RequestedAttribute Name="https://example.org/attributes/role"
      NameFormat=”urn:oasis:names:tc:SAML:2.0:attrname-format:uri”>
      <saml:AttributeValue>User</saml:AttributeValue>
      <saml:AttributeValue>Administrator</saml:AttributeValue>
    </md:RequestedAttribute>
</req-attr:RequestedAttributes>
  <vetuma xmlns="urn:vetuma:SAML:2.0:extensions">
    <LG>sv</LG>
  </vetuma>
</samlp:Extensions>`
}
```

README.md
kdhttps: thanks mate :)

cjbarth:
Do to some work splitting `node-saml` from `passport-saml`, you might want to continue to work on this over at [node-saml](https://github.com/node-saml/node-saml).

kdhttps:
@cjbarth  sure PR created https://github.com/node-saml/node-saml/pull/11

PR #612, opened 2021-06-17
Remove `node-saml` code and use an import instead

cjbarth:
This is a removal of all the SAML-specific code from this project and instead importing it from [node-saml](https://github.com/node-saml/node-saml). There are some changes needed to that repo to support this split, which are tracked [here](https://github.com/node-saml/node-saml/pull/9).

The idea is to have no functional changes with this PR as apparent from a consumer via the [Passport](https://www.passportjs.org/) framework, but there will still be major breaking changes to anyone that may have used the internals of this project for their own SAML purposes. This will be released as a 4.0.0 release when it is ready.


package.json
```diff
@@ -51,6 +51,7 @@
   },
   "dependencies": {
     "debug": "^4.3.1",
+    "node-saml": "file:../node-saml",
```
zoellner, line 54: this seems to expect a certain folder structure which is fine for local development during the transition but should probably at least be a import from a git hash before the PR lands (and quickly be changed to the actual npm package)

cjbarth: Absolutely. I wasn't going to land it like this. I wanted to get an NPM package in here; thus the other PR over there. I put this up so you could all see why I made the changes I did over on the node-saml project. If everything else looks good here, I'll adjust this reference as soon as node-saml is released as a package.

cjbarth:
I think this should be good to merge now that it references a beta version of `node-saml`. We can move to the real version once we are all sure of everything on both sides.

Landing this will allow me to make a PR for another bug fix with the SLO stuff.

zoellner (approved):

PR #615, opened 2021-06-24
restore BC for authnContext config

tgabi333:
As per the documentation this config allows to be a string or an array of strings

### DIFF
--- a/src/node-saml/types.ts
+++ b/src/node-saml/types.ts
@@ -98,7 +98,7 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
   acceptedClockSkewMs: number;
   attributeConsumingServiceIndex?: string;
   disableRequestedAuthnContext: boolean;
-  authnContext: string[];
+  authnContext: string | string[];
   forceAuthn: boolean;
   skipRequestCompression: boolean;
   authnRequestBinding?: string;


cjbarth:
After reviewing the code, it appears that the problem is the documentation, not the code. Have you tried `string` instead of `string[]` and gotten it to work?

tgabi333:
The previous versions supported this behavior. The generated javascript code did contained a transformation from string to string array in the constructor of this type.

tunderdomb:
@cjbarth I found the following investigating the same issue:

In #548 this behavior was removed from [saml.ts#154-156](
https://github.com/node-saml/passport-saml/pull/548/files#diff-864bd7929a1a1603afbae244cec33bac640a5bd41114620d085251f08f1ed352L154-L156):

```js
if (!Array.isArray(options.authnContext)) {
  options.authnContext = [options.authnContext];
}
```

The current code does not cover this use case in [saml.ts#152-154](https://github.com/node-saml/passport-saml/blob/master/src/node-saml/saml.ts#L152); if a string is passed, it is not converted to a string array:
```js
authnContext: ctorOptions.authnContext ?? [
  "urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport",
],
```

Even if the source file casts the value to a string array in [saml.ts#281](https://github.com/node-saml/passport-saml/blob/master/src/node-saml/saml.ts#L281):
```js
(this.options.authnContext as string[]).forEach(function (value) {
```

The compiled js output will be just:
```js
this.options.authnContext.forEach(function (value) {
```

So it appears that providing a string for this option is no longer supported; hence the breaking change.

Hope this helps!

cjbarth:
Yes, providing a string is no longer supported. So I'm not sure why we would want to add it back as a valid type. It appears that the documentation is now wrong.

Still, have you tried providing a `string` and gotten it to work? If not, then changing a type to say it would work would be wrong.

cjbarth:
Do to some work splitting `node-saml` from `passport-saml`, you might want to continue to work on this over at [node-saml](https://github.com/node-saml/node-saml).

Issue #419, opened 2020-02-02
passport-saml's IdP initiated LogoutRequest handling doesn't always close sessions


srd90:
It's possible that `passport-saml` responds to IdP initiated logout request that sessions are closed even though sessions are still alive. This can happen at least in following situation: 

1. end user has blocked third party cookies
1. SAML IdP is at another top level domain than SAML SP (passport-saml) application. Ie. SAML SP site is "third party site" from browsers' cookie handling point of view when browser is rendering page at SAML IdP site.
1. end user triggers global logout from some other service (passport-saml receives IdP initiated LogoutRequest)
1. SAML IdP has implemented logout propagation with combination of iframes and javascripts
  - https://wiki.shibboleth.net/confluence/display/IDP30/LogoutConfiguration#LogoutConfiguration-Overview
  - https://simplesamlphp.org/docs/stable/simplesamlphp-idp-more#section_1

Possible result from end use point of view: IdP reports that session related to `passport-saml` site is successfully closed even though it is not touched during logout process in anyway.

This problem has potential to hit a lot of end users when chrome starts to block third party cookies by default.

Following code is sending LogoutRequests to passport-saml like
- browser without third party cookie blocking enabled
- browser with third party cookie block enabled 

`console.logs` written by example code are available at the end of the issue.

```javascript
"use strict";

// Purpose: to demonstrate LogoutRequest handling issue.
// Drop this code to passport-saml's test/ directory and run tests.
// This is not meant to be added to passport-saml's testsuite as-is.
// This test code was "tested" against passport-saml version ac7939fc1f74c3a350cee99d68268de7391db41e
//
// Add following libraries to devDependencies
//   "supertest": "4.0.2",
//   "express-session": "1.17.0",
//   "cookie-parser": "1.4.4",
//   "chai": "4.2.0",
//   "chai-string": "1.5.0",
//   "memorystore": "1.6.1",
//   "cookiejar": "2.1.2"
//
//
//
const express = require("express");
const session = require("express-session");
const passport = require("passport");
const bodyParser = require("body-parser");
const SamlStrategy = require("../lib/passport-saml/index.js").Strategy;
const memorystore = require("memorystore")(session)

const supertest = require("supertest");
const chai = require("chai");
chai.use(require("chai-string"));
const expect = chai.expect;
const url = require("url");
const zlib = require("zlib");
const xmldom = require("xmldom");
const SignedXml = require("xml-crypto").SignedXml;
const fs = require("fs");
[...]

PR #619, opened 2021-07-01
Check user matches logout request before reporting logout success

cjbarth:
@srd90 Has reported a bug with SLO in node-saml, passport-saml, [here](https://github.com/node-saml/passport-saml/issues/419). This adjusts passport-saml to call node-saml correctly to get a success of fail response to send to the IdP. A logout is attempted no matter what, assuming that the SAML from the IdP is meant for us. However, a logout is only reported as successful if the user that is currently logged in matches the one that is specified in the logout request from the IdP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "Chris Barth",
     "Andrii Kostenko"
   ],
-  "main": "./lib/passport-saml",
+  "main": "./lib",
   "files": [
     "lib",
     "README.md",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "watch": "concurrently --kill-others \"npm:*-watch\""
   },
   "dependencies": {
-    "node-saml": "^4.0.0-beta.1",
+    "node-saml": "4.0.0-beta.1",
     "passport-strategy": "^1.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "express": "^4.17.1",
     "github-release-notes": "^0.17.3",
-    "mocha": "^9.1.1",
+    "mocha": "^9.1.3",
     "onchange": "^7.1.0",
     "passport": "^0.4.1",
     "prettier": "^2.4.1",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   },
   "dependencies": {
     "debug": "^4.3.1",
-    "node-saml": "^4.0.0-beta.0",
+    "node-saml": "file:../node-saml",
     "passport-strategy": "^1.0.0",
     "xml-crypto": "^2.1.2",
     "xml-encryption": "^1.2.4",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   },
   "dependencies": {
     "debug": "^4.3.1",
-    "node-saml": "file:../node-saml",
+    "node-saml": "^4.0.0-beta.1",
     "passport-strategy": "^1.0.0",
     "xml-crypto": "^2.1.2",
     "xml-encryption": "^1.2.4",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   "devDependencies": {
     "@types/debug": "^4.1.5",
     "@types/mocha": "^8.2.2",
-    "@types/node": "^14.14.44",
+    "@types/node": "^12.20.15",
     "@types/passport-strategy": "^0.2.35",
     "@types/request": "^2.48.5",
     "@types/sinon": "^10.0.0",
--- a/package.json
+++ b/package.json
@@ -50,26 +50,16 @@
     "watch": "concurrently --kill-others \"npm:*-watch\""
   },
   "dependencies": {
-    "debug": "^4.3.1",
     "node-saml": "^4.0.0-beta.1",
-    "passport-strategy": "^1.0.0",
-    "xml-crypto": "^2.1.2",
-    "xml-encryption": "^1.2.4",
-    "xml2js": "^0.4.23",
-    "xmlbuilder": "^15.1.1",
-    "xmldom": "^0.6.0"
+    "passport-strategy": "^1.0.0"
   },
   "devDependencies": {
-    "@types/debug": "^4.1.5",
-    "@types/mocha": "^8.2.2",
-    "@types/node": "^12.20.15",
+    "@types/mocha": "^9.0.0",
+    "@types/node": "^12.20.25",
     "@types/passport-strategy": "^0.2.35",
     "@types/request": "^2.48.5",
     "@types/sinon": "^10.0.0",
-    "@types/xml-crypto": "^1.4.1",
-    "@types/xml-encryption": "^1.2.0",
     "@types/xml2js": "^0.4.8",
-    "@types/xmldom": "^0.1.30",
     "@typescript-eslint/eslint-plugin": "^4.23.0",
     "@typescript-eslint/parser": "^4.23.0",
     "body-parser": "^1.19.0",
@@ -77,20 +67,21 @@
     "concurrently": "^6.1.0",
     "eslint": "^7.26.0",
     "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-prettier": "^3.4.0",
+    "eslint-plugin-prettier": "^4.0.0",
     "express": "^4.17.1",
     "github-release-notes": "^0.17.3",
-    "mocha": "^8.4.0",
+    "mocha": "^9.1.1",
     "onchange": "^7.1.0",
     "passport": "^0.4.1",
-    "prettier": "^2.3.0",
+    "prettier": "^2.4.1",
     "prettier-plugin-packagejson": "^2.2.11",
     "release-it": "^14.6.2",
     "request": "^2.83.0",
     "should": "^13.2.3",
-    "sinon": "^10.0.0",
-    "ts-node": "^9.1.1",
-    "typescript": "^4.2.4"
+    "sinon": "^11.1.2",
+    "ts-node": "^10.2.1",
+    "typescript": "^4.2.4",
+    "xml2js": "^0.4.23"
   },
   "engines": {
     "node": ">= 12"
--- a/src/strategy.ts
+++ b/src/strategy.ts
@@ -1,7 +1,7 @@
 import { Strategy as PassportStrategy } from "passport-strategy";
 import { strict as assert } from "assert";
 import * as url from "url";
-import { Profile, SAML, SamlConfig } from ".";
+import { Profile, SAML, SamlConfig, VerifiedCallback } from ".";
 import {
   AuthenticateOptions,
   RequestWithUser,
@@ -14,19 +14,28 @@ export abstract class AbstractStrategy extends PassportStrategy {
   static readonly newSamlProviderOnConstruct: boolean;
 
   name: string;
-  _verify: VerifyWithRequest | VerifyWithoutRequest;
+  _signonVerify: VerifyWithRequest | VerifyWithoutRequest;
+  _logoutVerify: VerifyWithRequest | VerifyWithoutRequest;
   _saml: SAML | undefined;
   _passReqToCallback?: boolean;
 
-  constructor(options: SamlConfig, verify: VerifyWithRequest);
-  constructor(options: SamlConfig, verify: VerifyWithoutRequest);
-  constructor(options: SamlConfig, verify: never) {
+  constructor(
+    options: SamlConfig,
+    signonVerify: VerifyWithRequest,
+    logoutVerify: VerifyWithRequest
+  );
+  constructor(
+    options: SamlConfig,
+    signonVerify: VerifyWithoutRequest,
+    logoutVerify: VerifyWithoutRequest
+  );
+  constructor(options: SamlConfig, signonVerify: never, logoutVerify: never) {
     super();
     if (typeof options === "function") {
       throw new Error("Mandatory SAML options missing");
     }
 
-    if (!verify || typeof verify != "function") {
+    if (!signonVerify || typeof signonVerify != "function") {
       throw new Error("SAML authentication strategy requires a verify function");
     }
 
@@ -38,7 +47,8 @@ export abstract class AbstractStrategy extends PassportStrategy {
       this.name = "saml";
     }
 
-    this._verify = verify;
+    this._signonVerify = signonVerify;
+    this._logoutVerify = logoutVerify;
     if ((this.constructor as typeof Strategy).newSamlProviderOnConstruct) {
       this._saml = new SAML(options);
     }
@@ -63,20 +73,15 @@ export abstract class AbstractStrategy extends PassportStrategy {
           // When logging out a user, use the consumer's `validate` function to check that
           // the `profile` associated with the logout request resolves to the same user
           // as the `profile` associated with the current session.
-          const verified = (err: Error | null, logoutUser?: User, info?: unknown) => {
+          const verified: VerifiedCallback = (err: Error | null, logoutUser?: User) => {
             if (err) {
               return this.error(err);
-            }
-
-            if (!logoutUser) {
-              return this.fail(info, 401);
             }
 
             let userMatch = true;
             try {
               // Check to see if we are logging out the user that is currently logged in to craft a proper IdP response
-
-              // TODO: This may be a problem if there are any timestamps in here, should we do a `strictEqual` instead?
+              // It is up to the caller to return the same `User` as we have currently recorded as logged in for a successful logout
 
               assert.deepStrictEqual(req.user, logoutUser);
             } catch (err) {
@@ -103,9 +108,9 @@ export abstract class AbstractStrategy extends PassportStrategy {
           };
 
           if (this._passReqToCallback) {
-            (this._verify as VerifyWithRequest)(req, profile, verified);
+            (this._logoutVerify as VerifyWithRequest)(req, profile, verified);
           } else {
-            (this._verify as VerifyWithoutRequest)(profile, verified);
+            (this._logoutVerify as VerifyWithoutRequest)(profile, verified);
           }
         } else {
           // If the `profile` object was null, this is just a logout acknowledgment, so we take no action
@@ -125,9 +130,9 @@ export abstract class AbstractStrategy extends PassportStrategy {
         };
 
         if (this._passReqToCallback) {
-          (this._verify as VerifyWithRequest)(req, profile, verified);
+          (this._signonVerify as VerifyWithRequest)(req, profile, verified);
         } else {
-          (this._verify as VerifyWithoutRequest)(profile, verified);
+          (this._signonVerify as VerifyWithoutRequest)(profile, verified);
         }
       }
     };
--- a/src/strategy.ts
+++ b/src/strategy.ts
@@ -183,7 +183,7 @@ export abstract class AbstractStrategy extends PassportStrategy {
               this.redirect(await this._saml.getAuthorizeUrlAsync(RelayState, host, options));
             }
           } catch (err) {
-            this.error(err);
+            this.error(err as Error);
           }
         },
         "logout-request": async () => {
@@ -198,7 +198,7 @@ export abstract class AbstractStrategy extends PassportStrategy {
               await this._saml.getLogoutUrlAsync(req.user as Profile, RelayState, options)
             );
           } catch (err) {
-            this.error(err);
+            this.error(err as Error);
           }
         },
       }[options.samlFallback];
--- a/src/strategy.ts
+++ b/src/strategy.ts
@@ -1,9 +1,11 @@
 import { Strategy as PassportStrategy } from "passport-strategy";
+import { strict as assert } from "assert";
 import * as url from "url";
 import { Profile, SAML, SamlConfig } from ".";
 import {
   AuthenticateOptions,
   RequestWithUser,
+  User,
   VerifyWithoutRequest,
   VerifyWithRequest,
 } from "./types";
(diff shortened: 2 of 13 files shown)


cjbarth:
@mikkopiu, @markstos I am @-mentioning you because of your involvement with the above-mentioned SLO issue.

cjbarth:
This is currently marked as a draft because it depends on the corresponding [PR](https://github.com/node-saml/node-saml/pull/10) in node-saml landing first and being released to NPM, at least as a beta, so this code can point to it.

src/strategy.ts
```diff
@@ -53,66 +53,83 @@ export abstract class AbstractStrategy extends PassportStrategy {
...
-        err: Error | null,
-        user?: Record<string, unknown>,
-        info?: Record<string, unknown>
-      ) => {
-        if (err) {
-          return this.error(err);
+          const RelayState = req.query?.RelayState || req.body?.RelayState;
+
+          if (
+            req.user != null &&
```
srd90, line 72: There is a `req.logout()` call at line 61.
IIRC it does two things (if there was `req.user` available)
1. clear `req.session.passport`
1. clear `req.user` which used to point to `req.session.passport.user`

so by the time this line gets executed `req.user` is always `null` and `Success` response would not be returned ever (which is safer response than current "return always `Success`" policy :) )?

src/strategy.ts
```diff
@@ -53,66 +53,83 @@ export abstract class AbstractStrategy extends PassportStrategy {
...
+            req.user.ID === profile.ID &&
```
srd90, line 73: Isn't this `ID` (e.g. `profile.ID`) just a random string associated to SAML protocol messages (and `profile.ID` would be random string associated to `LogoutRequest`)

src/strategy.ts
```diff
@@ -53,66 +53,83 @@ export abstract class AbstractStrategy extends PassportStrategy {
...
+            req.user.issuer === profile.issuer &&
+            req.user.nameID === profile.nameID &&
```
srd90, line 75: If developer configures SamlStrategy e.g. like this (example taken from `passport-saml`'s `README.md`):
```javascript
passport.use(new SamlStrategy(
  {
    path: '/login/callback',
    entryPoint: 'https://openidp.feide.no/simplesaml/saml2/idp/SSOService.php',
    issuer: 'passport-saml',
    cert: 'fake cert', // cert must be provided
  },
  function(profile, done) {
    findByEmail(profile.email, function(err, user) {
      if (err) {
        return done(err);
      }
      return done(null, user);
    });
  })
);
```
and lets assume that `findByEmail` returns e.g.
```
{
  userId: 42
}
```
value of `req.user` and `req.session.passport.user` would be that.

I.e. there would not be any any of the attributes (like `nameID`) which are now used to match authenticated user to content of `LogoutRequest`.

I was expecting to see comparison (dunno how it could be implemented) of just `nameID` and `sessionIndex` (and maybe issuer)

src/strategy.ts
cjbarth: I went back and read some source code and found your memory is correct. I've made adjustments to account for this.

src/strategy.ts
cjbarth: You are correct. Adjusted.

src/strategy.ts
cjbarth: This is really the sticky part. Some of the types obfuscated what really needed to be compared here. I've fixed the types and tried again. This time we're leveraging the consumers code to make sure we're comparing the correct thing. Thank you very much for your sharp eye. Hopefully the changes I've made make sense, however, there is still a potential problem with the method I've chosen (see the TODO). I'll resolve this thread so we can start a new thread based on the new code.

src/strategy.ts
```diff
@@ -57,38 +59,59 @@ export abstract class AbstractStrategy extends PassportStrategy {
...
+
+            if (!logoutUser) {
+              return this.fail(info, 401);
+            }
+
+            let userMatch = true;
+            try {
+              // Check to see if we are logging out the user that is currently logged in to craft a proper IdP response
+
+              // TODO: This may be a problem if there are any timestamps in here, should we do a `strictEqual` instead?
```
srd90, line 79: You commented at https://github.com/node-saml/passport-saml/pull/619#discussion_r663131078 :
>...there is still a potential problem with the method I've chosen (see the TODO). I'll resolve this thread so we can start a new thread based on the new code.

I don't have enough knowledge about JS/TS stuff to discuss about this TODO

cjbarth: It isn't so much about JS, it is about object comparison. Let's say that the `user` object returned by the consumer from a given `profile` object is something like this:

```js
{
  name: profile.nameID,
  lastLogIn: DateTime.now(),
}
```

A simple object like that makes the conversion from `profile` to `user` such that it would be different every time. That would break any chance of comparison and thus any chance of reporting a successful SLO.

The suggestion to do a `strictEqual` instead of a `strictDeepEqual` would only compare the top-level object properties, nothing nested. Thus, if there as a `datatime` property that is nested, that would not be used in the comparison. The assumption being that identity could be determined from the top-level properties. For example:

```js
{
  name: profile.nameID,
  loginData: {
    lastLogin: DateTime.now(),
    loginLocation: getLocationFromRequest(req)
  }
}
```

An object built like that would pass a `strictEqual`, but fail a `strictDeepEqual`.

Truthfully, it is almost looking like we'll need some other piece of information provided by the user for logout.

src/strategy.ts
```diff
@@ -89,16 +84,22 @@ export abstract class AbstractStrategy extends PassportStrategy {
...
+                profile,
+                RelayState,
+                options,
+                userMatch,
+                redirectIfSuccess
               );
             }
+
+            // Log out the current user no matter if we can verify the logged in user === logout requested user
+            req.logout();
```
srd90, line 102: I'm not sure about control flow but based on quick code browsing I think that if implementation of `redirectIfSuccess` callback https://github.com/node-saml/passport-saml/blob/591a1d83b8ebfb2c19a2e393b37238974687d090/src/strategy.ts#L135-L143 gets executed by implementation of `getLogoutResponseUrl` this `req.logout()` line is not reached and `passport-saml` would once again return `Success` even if session was not terminated. Although this time reason would be that
1. even though caller was identified - i.e. his/her browser didn't block 3rd party cookie - and
2. this PR's code matched `req.user` with `LogoutRequest`'s information meaning that it was determined that it is a-ok to return `Success` but
3. redirect with `LogoutResponse` with `Success` status was triggered and request processing ended  before reaching this line

Your pull request seems to contain testcases i.e. TDD approach was your goal. I haven't inspected test case changes but based on already closed review findings (issues weren't captured by testcases) it seems that there aren't any tests which would replicate situation from end user's UE point of view / signalling between end user's UE and express.js application at SAML protocol level / HTTP request cookie / actual resource authorization level.

I attached to #419 's description two testcases which try to verify (and at that particular case pin point issues) `passport-saml`'s behaviour at SAML/HTTP protocol level along with enablers for session management tests. 

BTW. given the various possibilities to configure express-session to e.g. update session content after request processing to session store and various session store implementations is there any chance to make a decision - by passport-saml - that content of `session.passport` for given `sessionID` was successfully cleared from session store because:
Assume that
1. developer has configured express session to write session back to session store after each  request
2. application is some SPA application running at end user's browser tab
3. SPA application uses e.g. long polling to get some updates from backend
4. end user triggers SLO from some application B (running at another tab or browser window) which uses same SSO with aforementioned SPA application
5. IdP coordinates `LogoutRequest` to aforementioned SPA application's backend (this coordination happens at tab/browser window used by application B)
6. SPA application had ongoing long polling (which started prior to SLO and ends after SLO has ended)

now...if express session is configured to always update `session` to session store after `req` is processed doesn't that mean that express/nodejs instance which was handling concurrently aforementioned long polling request write content of `session` (along with `session.passport.user`) from `req` which represented long polling request back to session store. I.e. removed session would be reintroduced to session store?

If that concurrent request handling scenario is possible then `passport-saml` would have to have its own book keeping of login sessions (i.e. session would be added to bookkeeping only once during successfull authentication) and removed on normal logout or SLO `LogoutRequest` processing and there would not be any chance to mess things up with uncontrollable/wild 3rd party sw/configurations...and passport-saml's own bookkeeping would need to be checked during each request if request is related to resource which must not be accessible for unauthenticated users.

Sorry if it was hard to interpret this comment (grammar etc.). Due to some reason it was harder than usual to write english.

src/strategy.ts
```diff
@@ -57,38 +59,59 @@ export abstract class AbstractStrategy extends PassportStrategy {
...
         if (profile != null) {
           if (this._saml == null) {
             throw new Error("Can't get logout response URL without a SAML provider defined.");
           }
 
-          const RelayState = req.query?.RelayState || req.body?.RelayState;
+          const getLogoutResponseUrl = this._saml.getLogoutResponseUrl;
+          // When logging out a user, use the consumer's `validate` function to check that
+          // the `profile` associated with the logout request resolves to the same user
+          // as the `profile` associated with the current session.
```
srd90, line 65: Come to think of it this approach https://github.com/node-saml/passport-saml/blob/80fc3a2230874f8698c3e03a27818041a4d967fb/src/strategy.ts#L68-L70 would not work if verify function - like the one given at `passport-saml`'s [README.md](https://github.com/node-saml/passport-saml/blob/bba77bcc8030c633a2df5dd25f5c86f6caef5ce8/README.md) - uses some other information (assertion) from SAML login response than `nameID` to lookup user e.g. from DB.

`LogoutRequest` contains only `nameID` and `sessionIndex` thus `profile` constructed from `LogoutRequest` would not contain e.g. `email` information.

I.e. if `verify` function would be
```javascript
  function(profile, done) {
    findByEmail(profile.email, function(err, user) {
      if (err) {
        return done(err);
      }
      return done(null, user);
    });
  }
```
it would not resolve to same `user` object with the one that was resolved during login phase `verify` function processing.

srd90:
Based on discussions at #419 and in this particular PR and implementation of https://github.com/espoon-voltti/evaka/pull/738 it should be clear that there aren't any silver bullets which could be used to fix `passport-saml` so that it would return at least proper status code at `LogoutResponse` let alone to fix it so that IdP initiated LogoutRequest would just work "out of the box" from client programs point of view.

One solution would be to disable SLO feature by default and let developers to explicitly enable it if and when they understand the very specific conditions when `node-saml/passport-saml`'s SLO current implementation has any chance to work proberly (i.e. if IdP and SP are at same TLD and TLD is not at public suffix list(*)  etc.).

SLO feature flag would
1. increase awareness of `passport-saml`'s SLO flaw. This problem is not going to fade away.  It is easy to ignore this SLO flaw as long as one doesn't have to make explicit decisions whether to use it (current SLO implementation) or not regardless of known issues
2. introduce incentive to create and share solution for it i.e. force client applications to take charge of using flawed SLO implementation or to create fix to it (and share that fix with rest of the community if they don't want to keep maintaining it privately)

Another solution to increase awareness and create incentive to come up with solution of/to SLO weakness would be to create e.g Github Security Advisory (https://docs.github.com/en/code-security/security-advisories/creating-a-security-advisory) and request CVE number for it. This way issue would become more visible (e.g. npm might/could raise a warning about it) and once again client programs using `pasport-saml` would have to evaluate and take proper actions to handle that security warning.

(*)
https://wiki.mozilla.org/Public_Suffix_List
https://publicsuffix.org/
https://publicsuffix.org/list/
( links contained valid information/were referenced 3th of July 2021 )

@cjbarth @markstos 

btw. that not so long ago fixed "missing cert configuration option allows authentication bypass" issue would be also a good candidate for GSA

cjbarth:
I feel like since `nameID` is required in the SAML profile, that would be how one would have to look up a user in a database, or whatever other method is used to store sessions. In any case, if we make this change to enforce a deep equal of the user object, and that doesn't happen, SLO is effectively disabled (the default becomes `fail` instead of `success`). If a deep equal works, then I think we can be confident that we are logging out the correct user.

I'll see about creating a CVE.

srd90:
@cjbarth you commented:
> I feel like since `nameID` is required in the SAML profile, that would be how one would have to look up a user in a database, or whatever other method is used to store sessions.

About nameids. There are different nameid formats and especially transient. See https://stackoverflow.com/questions/11693297/what-are-the-different-nameid-format-used-for

About transient and Shibboleth:
>…
>Strictly speaking, SAML assertions don't have to contain a name identifier. The subject may be implicitly identified as the bearer of the token or anybody able to demonstrate possession of a key. In SSO use cases, one reason for including an identifier is to enable the relying party to refer to the subject later, such as in a query, or a logout request. So-called "transient" identifiers that are generated uniquely for each assertion are often used to support those use cases and are a common pattern in Shibboleth deployments.
>…
> Shibboleth deployments traditionally have focused on the use of Attributes to describe subjects, and default to the use of transient name identifiers (or omitting them). Commercial SAML deployments less commonly make use of Attributes and tend to use loosely or improperly specified name identifiers.
>…

source: https://wiki.shibboleth.net/confluence/display/CONCEPT/NameIdentifiers

So there are quite a few (Shibboleth) IdPs which communicate identity of the user with some attribute (i.e. IdP and SP are aware which attribute that is) instead of nameid.

So there are quite a few `verify` function implementations out there which expect to find some other attribute than value of NameID from the `profile` provided by `passport-saml` in order to lookup user information from some datastore (see passport samls own example of verify function which uses email attribute to fetch user data). If application defined `verify` functions use case is changed so that it is suddenly (from application pov) used also during logout phase and with `profile` without neccessary information the result would quite probably be some error condition instead of clean ”logout success/failure response”.

@cjbarth you wrote:
>  If a deep equal works, then I think we can be confident that we are logging out the correct user.

Unless session is reintroduced into session store due combination of some resave funtionality: https://github.com/expressjs/session/tree/0048bcac451ad867299d404aca94c79cc8bc751d#resave
and concurrent request processing see ”BTW.” from this comment: https://github.com/node-saml/passport-saml/pull/619#discussion_r663201340

What I am trying to say is that passport-saml cannot be confident unless it is  involved in authorization loop of each request or unless it delegate whole processing of logout to application.

src/strategy.ts
cjbarth: I agree with your scenario, however, the assumption in (1) is really not passport-saml's problem. The developer shouldn't be adding things to a user store just because a request completed.

Since it is out of scope for passport-saml to manage a user store, there has to be some onus on the developer to do this correctly. While there may be opportunities to use passport-saml incorrectly, my goal here is to make sure that it doesn't do that _by default_ as your previous bug reported pointed out.

src/strategy.ts
cjbarth: You are entirely correct. Since we don't currently have anyone running into this problem, I'm tempted to gloss over this until we can get some real-world use-cases. There are things that can be done with the IdP to help this matter, like including more data on SLO requests and duck-typing of `profile` can be used to branch this function internally for login/logout.

cjbarth:
> What I am trying to say is that passport-saml cannot be confident unless it is involved in authorization loop of each request or unless it delegate whole processing of logout to application.

I think I've come to the same understanding you have. The only method that is truly in-scope for passport-saml is to delegate this to the application. See https://github.com/node-saml/passport-saml/pull/619#discussion_r712246505. It would be a braking change for sure, but at least it would call attention to the fact that a developer actually needs to think about their SLO flows, whereas right now it appears automatic to the developer.

src/strategy.ts
cjbarth: There is now a second verify function so the user can do the calculation on finding the user to log out. If they return the same object that we have referenced internally, then we consider the logout successful, if not, then we consider the logout a failure.

Given there is a unique function for logout, the user can then cause whatever credential store, or whatever else they have, to do a logout as well and even handle their own sync issues.

Do you see any problems with this method @srd90 @markstos ?

src/strategy.ts
cjbarth: I've implemented a second verify function to allow the user to keep the implementations separate. They will have to actually do the calculations to figure out which user is to be logged out.

package.json
```diff
@@ -51,7 +51,7 @@
   },
   "dependencies": {
     "debug": "^4.3.1",
-    "node-saml": "^4.0.0-beta.0",
+    "node-saml": "file:../node-saml",
```
srd90, line 54: btw. is this change correct? there isn't `../node-saml` in the source tree (following link points to source tree of the [commit](https://github.com/node-saml/passport-saml/pull/619/commits/df81d0f5651a5e915e1fa27ce16c12122f33d8cd) which changed this line: https://github.com/node-saml/passport-saml/tree/df81d0f5651a5e915e1fa27ce16c12122f33d8cd/src )

src/strategy.ts
srd90: (Just to make it easier for those who might read this conversation later @cjbarth is referring to this modification [cade87c](https://github.com/node-saml/passport-saml/pull/619/commits/cade87c5a73f27561da036fd9101549f620b8518))

https://github.com/node-saml/passport-saml/blob/cade87c5a73f27561da036fd9101549f620b8518/src/strategy.ts#L86 check works if and only if express/passport was able to associate correct user to `req.user` in the first place. I.e. if there was proper session cookie along with the incoming `LogoutRequest` and thus `req.isAuthenticated` (or similar) is/would/could also return `true` and in this particular case this check would double check that authenticated session was created from login response impliciated by content of logoutrequest.

If session cookie was missing that check would fail always unless developer circumvents it by injecting object `foo` to `req.user` during this invocation https://github.com/node-saml/passport-saml/blob/cade87c5a73f27561da036fd9101549f620b8518/src/strategy.ts#L111 and then return same `foo` as a result of `VerifyWithRequest` callback invocation.

IMHO with [cade87c](https://github.com/node-saml/passport-saml/pull/619/commits/cade87c5a73f27561da036fd9101549f620b8518) approach `passport-saml` could return LogoutResponse with failure status without even consulting veifylogout callback if `req.user` was null or `req.isAuthenticated !== true`.

On the other hand current implementation would make it possible for above average developer to implement SAML compliant session management (i.e. session management which would support logging out without session cookie) because now it would be possible to have application specific code to be executed during LogoutRequest processing. Only problem would be that `passport-saml` uses `req.user` to determine whether it - `passport-saml` - should respond with Success to LogoutRequest. Workaround would  be to inject `foo` to `req.user` and return `foo` as a result of verifylogout callback. Not too clean API from above average developer point of view (and those who just don't care about correct SLO implementation would use same workaround to have logoutresponse with success status).

How about if verifylogout callback function would take same parameters as it takes [at the moment](https://github.com/node-saml/passport-saml/pull/619/commits/cade87c5a73f27561da036fd9101549f620b8518) but change its name to `executeSLOLogout` and if the implementation of that callback function returns a string `"I as a developer of this function solemnly swear and affirm that the session implicated by LogoutRequest is successfully terminated and it - session - is not reintroduced to session store even if there was concurrent request processing ongoing wich handled same user while LogoutRequest was processed"` (or maybe just `true`) passport-saml would respond with logoutresponse with `Success`. `passport-saml` could call behind the scenes req.logout() just before it sends logoutresponse but that call would be sort of "last effort". My point is that API could be nicer for those who are interested to implement proper IdP initiated SLO.

btw. now I remembered seeing this https://github.com/node-saml/passport-saml/pull/144 which is somewhat related to this pull request's discussion (possibility to execute custom logic after LogoutRequest signature verification but before response delivery). First answer in that pull request didn't address pull request author's second concern (idp initiated case) at all. I linked to that pull request just in case someone is interested browse related discussions.

srd90: I interpreted your comments https://github.com/node-saml/passport-saml/pull/619#issuecomment-923005584 and https://github.com/node-saml/passport-saml/pull/619#discussion_r712246505 so that you planned to drop "deepStrictEqual" check (quote from latter linked comment):
>The more I think about it, the more I think that is the only way we can really let the user make the call. Then we'd just pass the user that we have and the profile that we got from the SLO request, and let the user of passport-saml return a go/no-go on the validity of the SLO request instead of trying to do that calculation blind in the passport-saml code.

package.json
cjbarth: This change is only temporary in order to test changes. We won't land the code like this. When this is completely approved and ready to land, we'll land the `node-saml` part first, then publish it, and then point this to that.

src/strategy.ts
cjbarth: I see your point about a potentially fuzzy API here. My thinking was that we want to guarantee that the user that we're working with is the one we are telling the IdP that we're logging out. I don't want the consumer to be able to write a function like `() => true` and have SLO work. By forcing them to actually go get a `user` object (preferably from their user store, but possibly `() => req.user`, which should be a red flag to any programmer) and return it, and then we make sure it is 100% the same (so deep equal), we have strong assurances that we aren't lying to the IdP.

If there is no match, we still log out any current user, so that covers all the other cases, all without lying to the IdP.

I can see renaming the function, in fact, both of the functions aren't really a "verify"... but I'm not so sure they are "execute" either.

markstos (changes requested):
The proposal here to verify the user doing logout is sound, but it appears that the PR could use a rebase or perhaps even a fresh start in a new branch. 

The current PR adds over 1,000 lines and removes over 12,000. Surely that kind of churn isn't required to verify a user during logout.

cjbarth:
Acknowledged @markstos . I've split off some of the changes here that aren't strictly related to logout validation. That should help thin out this PR a little once that lands and I rebase this one.

https://github.com/node-saml/passport-saml/pull/658

markstos:
Thanks, @cjbarth. It's looking better.

markstos (approved):

src/strategy.ts
cjbarth: @markstos , since you've approved this PR, I'm assuming your up to speed on this discussion and favor leaving these functions named "verify", so I'll land this PR. If there is a need to revisit, I'm more than open to making that simple refactor before we leave beta.

PR #633, opened 2021-08-24
Update xmldom to 0.7.2 - branch 3.x

forty:
The package is now scoped under @xmldom. See https://github.com/xmldom/xmldom/pull/278
This fixes security vulnerability CVE-2021-32796.
Also update xml-crypto which depends on xmldom.

This is the same as https://github.com/node-saml/node-saml/pull/17 for the 3.x branch which should be easier to release quickly.


forty:
Updated the PR to update xml-encryption as well, has they merged to xmldom update

PR #641, opened 2021-09-20
Update packages

cjbarth:
Packages were removed where possible and otherwise updated. The only package that was downgraded was the types package for Node. It was at v14, but we support Node v12 and up, so I change the types to be Node v12-compatible. We can update this package as we deprecate older versions of Node.


markstos (approved):
I presume we can remove all these deps now due to the node-saml split. Looks good to me!

Issue #642, opened 2021-09-21
[BUG] Wrong readme example for authnContext for ADFS


jyenduri-uptycs:
** Spec-driven development **

This project is focused on compliance with the SAML 2.0 specification. For any bug report that
involves the SAML spec, please link to the related parts of the spec and quote the passages too.

Start here: http://saml.xml.org/saml-specifications

You might also check the spec to confirm that it doesn't address your particular bug and mention
that you found no references in the spec concerning your issue.

** Community development model **

passport-saml is maintained by a number of current users. There is no author or primary maintainer
waiting to write your tests and documentation for you. To increase the odds that your issue
is promptly dealt with, consider a pull request to address the issue that includes test coverage
and updated documentation.

**To Reproduce**

I am trying out ADFS  based on the example `authContext` is a string, but looks like it is only working if we give it as an array

{
    entryPoint: 'https://ad.example.net/adfs/ls/',
    issuer: 'https://your-app.example.net/login/callback',
    callbackUrl: 'https://your-app.example.net/login/callback',
    cert: 'MIICizCCAfQCCQCY8tKaMc0BMjANBgkqh ... W==',
    authnContext: 'http://schemas.microsoft.com/ws/2008/06/identity/authenticationmethod/windows',
    identifierFormat: null
  }

Working example

{
    entryPoint: 'https://ad.example.net/adfs/ls/',
    issuer: 'https://your-app.example.net/login/callback',
    callbackUrl: 'https://your-app.example.net/login/callback',
    cert: 'MIICizCCAfQCCQCY8tKaMc0BMjANBgkqh ... W==',
    authnContext: ['http://schemas.microsoft.com/ws/2008/06/identity/authenticationmethod/windows'],
    identifierFormat: null
  }

**Expected behavior**
We need to update the README and examples

**Environment**

- Node.js version: v12.14.0
- passport-saml version: 3.1.2

jyenduri-uptycs:
@cjbarth  Can you please look at this pull request https://github.com/node-saml/passport-saml/pull/647

ctaschereau:
Agreed. This appears to be an undocumented breaking change with the version 2.x since we had working code passing only a string as an authncontext before...

cjbarth:
I found the commit that created the break. It was my [commit](https://github.com/node-saml/passport-saml/pull/548), and I overlooked updating the documentation. Thanks for catching this.

PR #647, opened 2021-09-21
Read me update for authnContext example for ADFS

jyenduri-uptycs:
# Description
Update the readme document for ADFS example

Bug submitted https://github.com/node-saml/passport-saml/issues/642

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ Here is a configuration that has been proven to work with ADFS:
     issuer: 'https://your-app.example.net/login/callback',
     callbackUrl: 'https://your-app.example.net/login/callback',
     cert: 'MIICizCCAfQCCQCY8tKaMc0BMjANBgkqh ... W==',
-    authnContext: 'http://schemas.microsoft.com/ws/2008/06/identity/authenticationmethod/windows',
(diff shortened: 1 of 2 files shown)


cjbarth (approved):

cjbarth:
Please run `npm run prettier-format` and then push. I'll do a squash merge, so it will be simple.

jyenduri-uptycs:
@cjbarth  

```
npm run test
```
Above script works fine in my local. Is there anything I should fix?

cjbarth (dismissed):

cjbarth:
I found the problem. It is because our build looks for upcoming potential problems by updating all packages as part of the build. In this case, it uncovered an issue with out code related to an update to TypeScript type checking. This problem is addressed in https://github.com/node-saml/passport-saml/pull/641. When that lands, then I'll re-run the build jobs here and this will land just fine.

Thanks for your contribution!

jyenduri-uptycs:
Perfect! Thanks @cjbarth

PR #654, opened 2021-10-06
Abort option to callback

ovictoraurelio:
**Summary**

I have added a third parameter (_abort_) to done callback in getSamlOptions.

**Addressed**

In some cases I need to avoid or "pass" SAML authentication because doesn't exists any SAML Configuration for a Specific Tennant of my multi-tennancy, multi-saml and multi passport strategies application.

We are adding support to integrate authentication system of our App with SAML to our clients (Each one is a Tennant). So our middleware of authentication includes auth strategy of SAML and the getSamlOptions sounds the perfect place to find out if this specific Tennant has a valid SAML Configuration to continue to SAML authorization otherwise we just need to "abort" the SAML authentication making then passport will try another authentication method.

 **Checklist**:

- Issue Addressed: [ YES ]
- Link to SAML spec: [ N ]
- Tests included? [ N ]
- Documentation updated? [ YES ]

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ passport.use(new MultiSamlStrategy(
         if (err) {
           return done(err);
         }
+        if (!provider) {
+          return done(null, {}, true)
+        }
         return done(null, provider.configuration);
       });
     }


ovictoraurelio:
@cjbarth When I returned an error or an empty payload passport will not try another authentication method and the endpoint will just return a error instead of continue execution.

ovictoraurelio:
Hi @cjbarth  first of all, thanks a lot of taking care about my case.

I'll paste my entire passport-saml code below to be more clear.

```javascript
import logger from '@/logger'
import { MultiSamlStrategy } from 'passport-saml'
import qs from 'qs'
import { getUserFromAuthCredentials } from './utils'
import { Op } from 'sequelize'

/**
 *
 * @param {{cacheController: import('@/config/CacheController').default}} param0
 * @returns
 */
export function multiSamlStrategy ({ cacheController }) {
  return new MultiSamlStrategy(
    {
      passReqToCallback: true,
      // failureFlash: true,
      cacheProvider: cacheController,
      /**
       *
       * @param {import('express').Request} req
       * @param {Function} done
       */
      getSamlOptions: async function (req, done) {
        try {
          const { db } = req
          const additionalParams = req?.body?.RelayState ? JSON.parse(req?.body?.RelayState) : req.query

          const { data } = await db.Credentials.findOne({
            attributes: ['data'],
            include: [{
              model: db.ConfigLoginMethods,
              required: true,
              attributes: ['id'],
              where: {
                loginMethodId: additionalParams.loginMethodId
              }
            }]
          })

          if (!data) {
            // I want to abort!!
            // calling done(null, null, true) will catch a error: "cert is required"
            // doing nothing.. will just make the request keep waiting and don't call express... next()
            // return null will catch a error: "cert is required"
          }

          if (!data.options) {
            data.options = {}
          }

          data.callbackUrl = data.callbackUrl.concat(`?${qs.stringify(additionalParams)}`)
          data.additionalParams = {
            RelayState: JSON.stringify(additionalParams)
          }

          done(null, data)
        } catch (err) {
          done(err)
        }
      }
    },
    async function (req, profile, done) {
      try {
        if (req?.user?.uuid) {
          logger.info(`user: ${req.user.uuid} restored from session`)
          return done(null, req.user)
        }

        const db = req.db
        const { uuid, user } = await getUserFromAuthCredentials(db, req.params.orgSlug, { unsafe: true, email: profile.nameID })
        const userData = user

        logger.profile('MultiSamlStrategy', { level: 'verbose' })

        if (process.env.REDIS_HOST) {
          cacheController.addUserSession({ uuid, userData })
        }

        done(null, {
          ...userData,
          passportStrategy: 'MultiSamlStrategy'
        })
      } catch (err) {
        if (err.status === 401) {
          done(null, false, { message: err.message })
        } else {
          done(err)
        }
      }
    })
}
```

My function to getSamlOptions will catch all SAML SSO Setup infos, cause in my application the informations about SSO Configs will be stored in a Database and KMS.
Each tennant of my application could have none, one or multiple SSO SAML Setups.

Sorry about my english mistakes...

PR #660, opened 2021-12-15
Document passive option

sherakama:
# Description

Document support for passive option in `node-saml`

Re: https://github.com/node-saml/node-saml/blob/master/src/saml.ts#L135

# Checklist:

- Issue Addressed: []
- Link to SAML spec: []
- Tests included? []
- Documentation updated? [x]

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ export interface Profile {
 - `racComparison`: Requested Authentication Context comparison type. Possible values are 'exact','minimum','maximum','better'. Default is 'exact'.
 
 - `forceAuthn`: if set to true, the initial SAML request from the service provider specifies that the IdP should force re-authentication of the user, even if they possess a valid session.
-- `passive`: if set to true, the initial SAML request from the service provider specifies that the IdP should prevent visible user interaction. The IdP recognizes the parameter and returns to the SP. This is useful for validating a user session. 
+- `passive`: if set to true, the initial SAML request from the service provider specifies that the IdP should prevent visible user interaction. This is useful for validating a user session without prompting for a login when there is no active session. The IdP recognizes the parameter and returns to the SP. 
   - An error if the IdP must interact with the user but cannot because of this parameter.
   - A Federation Assertion that indicates whether the user has a valid session. 
 - `providerName`: optional human-readable name of the requester for use by the presenter's user agent or the identity provider
--- a/README.md
+++ b/README.md
@@ -165,9 +165,9 @@ export interface Profile {
 - `racComparison`: Requested Authentication Context comparison type. Possible values are 'exact','minimum','maximum','better'. Default is 'exact'.
 
 - `forceAuthn`: if set to true, the initial SAML request from the service provider specifies that the IdP should force re-authentication of the user, even if they possess a valid session.
-- `passive`: if set to true, the initial SAML request from the service provider specifies that the IdP should prevent visible user interaction. This is useful for validating a user session without prompting for a login when there is no active session. The IdP recognizes the parameter and returns to the SP. 
+- `passive`: if set to true, the initial SAML request from the service provider specifies that the IdP should prevent visible user interaction. This is useful for validating a user session without prompting for a login when there is no active session. The IdP recognizes the parameter and returns to the SP.
   - An error if the IdP must interact with the user but cannot because of this parameter.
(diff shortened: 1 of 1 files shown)


cjbarth:
Thanks for adding that documentation. Do you happen to have a reference for the part of the SAML spec that refers to? That feature was added a while ago by someone else. I'd like to refresh my understanding of your description before approving :)

sherakama:
There is this https://docs.oracle.com/en/middleware/idm/access-manager/12.2.1.3/aiaag/introducing-identity-federation-oracle-access-management.html

I'll see about something more specific to the SAML 2.0 spec.

sherakama:
https://www.oasis-open.org/committees/download.php/8958/sstc-saml-implementation-guidelines-draft-01.pdf

Page 33 - Section 7.2.1

cjbarth:
I'm having trouble finding actual documentation on the `IsPassive` flag that `passport-saml` actually sets. The best I've found so far discusses largely what you've said, but still no references to it being part of a spec: https://lists.oasis-open.org/archives/saml-dev/201605/msg00004.html


> Hi,
> 
> theoretically such seemless authentication is still within the bounds of the isPassive definition. Such authentication methods could probably involve: * kerberos (as long as the 401 page is not being displayed to the end-user after failed login)
> 
> * remember me feature (a.k.a persistent cookie)
> 
> Cert based authn is also similar, but with that the browser will ask for the password of the private key, not sure if that qualifies as passive authn.
> 
> 
> cheers,
> Peter
> 
> 2016. 05. 10. 12:17 keltezéssel, Chiranga Alwis írta:
> 
> > Hi Dieter,
> > 
> > first of all thanks for the answer.
> > 
> > In addition, to what you have mentioned I came across this alternative
> > usage of this property (from another source) which enables the Service
> > Provider to ask the IdP to basically "authenticate this user only if you
> > can do it without the having the user involved". This seems to map with
> > the spec definition provided for the 'IsPassive' property under
> > discussion. Is this a valid usage and what does it accurately mean in a
> > real-world scenario?
> > 
> > 
> > 
> > > On Tue, May 10, 2016 at 1:18 PM, Dieter Maurer <dieter@handshake.de

README.md
```diff
@@ -165,6 +165,7 @@ export interface Profile {
 - `racComparison`: Requested Authentication Context comparison type. Possible values are 'exact','minimum','maximum','better'. Default is 'exact'.
 
 - `forceAuthn`: if set to true, the initial SAML request from the service provider specifies that the IdP should force re-authentication of the user, even if they possess a valid session.
+- `passive`: if set to true, the initial SAML request from the service provider specifies that the IdP should prevent visible user interaction if possible.
```
cjbarth, line 168: Based on the spec, "if possible" isn't true. The spec says the IdP "must not". Is there a different resource where you found that it obeys this less strict implementation?

cjbarth:
Here it is on page 49: https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf

> ForceAuthn [Optional]
> A Boolean value. If "true", the identity provider MUST authenticate the presenter directly rather than
> rely on a previous security context. If a value is not provided, the default is "false". However, if both
> ForceAuthn and IsPassive are "true", the identity provider MUST NOT freshly authenticate the
> presenter unless the constraints of IsPassive can be met.
> IsPassive [Optional]
> A Boolean value. If "true", the identity provider and the user agent itself MUST NOT visibly take control
> of the user interface from the requester and interact with the presenter in a noticeable fashion. If a
> value is not provided, the default is "false".

README.md
sherakama: > Based on the spec, "if possible" isn't true. The spec says the IdP "must not". Is there a different resource where you found that it obeys this less strict implementation?

I can't say I have for SAML. I am happy to change this language.

sherakama:
@cjbarth,

Thanks for reviewing this and finding the documentation. I have updated the language. I hope this helps.

cjbarth (dismissed):

cjbarth:
Please run `npm run prettier-format` to correct the code formatting.

sherakama:
Ok.

cjbarth (approved):

sherakama:
Thanks @cjbarth

PR #667, opened 2022-01-19
Update xml-encryption to get rid of vulnerable node-forge

forty:
This update actually /remove/ node forge to use native nodejs crypto instead. This is nice as node-forge is a fairly large module.

The major version bump in xml-encryption is because this update remove support for node < 12 (which are already not supported by passport-saml)

This PR is on the 3.x branch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,14 +5,15 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


markstos:
@forty Thanks for the details. Merging.

Issue #674, opened 2022-02-23
[Bug] passport-saml is not exporting enough types so we can segregate the code easily


thejaigoyal:
## To Reproduce

```
npm i passport-saml
```

```
import { MultiSamlConfig } from 'passport-saml';

const multiSamlOptions: MultiSamlConfig = {
  passReqToCallback: true, // makes req available in callback
  getSamlOptions: async (request, done) => {
    const configuration = await findProvider(request);
    return done(null, configuration);
  },
};

const multiSamlStrategy = new MultiSamlStrategy(multiSamlOptions, login);
```

importing MultiSamlConfig throws below error
`Module '"passport-saml"' has no exported member 'MultiSamlConfig'.ts(2305)`

thejaigoyal:
PR #675  has been created!

PR #675, opened 2022-02-23
🎨  add MultiStrategyConfig type export

thejaigoyal:
# Description
Exporting `MultiStrategyConfig` type from src index so the type is available and can be imported for segregating the code and be used like below

```
import { MultiSamlStrategy, VerifyWithRequest, MultiSamlConfig } from 'passport-saml';

const multiSamlOptions: MultiSamlConfig = {
  passReqToCallback: true, // makes req available in callback
  getSamlOptions: async (request, done) => {
    const configuration = await findProvider(request);
    return done(null, configuration);
  },
};

export const multiSamlStrategy = new MultiSamlStrategy(multiSamlOptions, login);
```

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,12 @@
 import { Strategy, AbstractStrategy } from "./strategy";
 import { MultiSamlStrategy } from "./multiSamlStrategy";
 
-import type { VerifiedCallback, VerifyWithRequest, VerifyWithoutRequest } from "./types";
+import type {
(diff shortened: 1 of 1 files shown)


thejaigoyal:
This P belongs to Issue #674

cjbarth (approved):

Issue #549, opened 2021-03-04
[BUG] TypeScript Strategy is incompatible with Passport Strategy


fluggo:
Small bug: When using the types from the latest passport-saml, a new Strategy object can't be used in a passport.use() call.

**To Reproduce**

```typescript
passport.use(new passportSaml.Strategy({} as any, (profile: any, done: passportSaml.VerifiedCallback) => {
  done(null, profile, {});
});
```

**Expected behavior**
The docs suggest this very usage, so it should pass TypeScript checks without errors. Unfortunately, the strategy's authenticate() method requires a [samlLogoutRequest](https://github.com/node-saml/passport-saml/blob/master/src/passport-saml/types.ts#L146) property to appear on the request object, and since express doesn't provide that, it fails.

**Environment**

- Node.js version: 14.15.4
- TypeScript version: 4.1.5
- passport-saml version: 2.0.5

cjbarth:
@fluggo What docs suggest this? I'm not sure what you're trying to accomplish here. Also, it seems you're missing a `)`, as there aren't enough, so the syntax isn't correct. Sorry if I'm missing something obvious.

fluggo:
Ah, you're right about the parentheses. Sorry, I was in a hurry, and was adapting my code. Let's try with a complete TypeScript file, this time directly from the main readme:

```typescript
import * as passport from 'passport';
import { Strategy as SamlStrategy } from 'passport-saml';

passport.use(new SamlStrategy(
  {
    path: '/login/callback',
    entryPoint: 'https://openidp.feide.no/simplesaml/saml2/idp/SSOService.php',
    issuer: 'passport-saml'
  },
  function(profile, done) {
    findByEmail(profile.email, function(err, user) {
      if (err) {
        return done(err);
      }
      return done(null, user);
    });
  })
);
```

TypeScript gives this error:

![image](https://user-images.githubusercontent.com/2552507/110062335-b135e900-7d2e-11eb-9f4d-9333d0d314e9.png)

cjbarth:
@fluggo I still can't seem to duplicate this problem. I've checked out `passport-saml` at `v2.0.5` and `npm install --save-dev @types/passport@1.0.6` and then put the following code in `passport-saml/src/test.ts` with the following code:

```js
import * as passport from "passport";
import { Strategy as SamlStrategy } from "./passport-saml";
import { Profile, VerifiedCallback } from "./passport-saml/types";

[...]

Issue #221, opened 2017-07-03
IdP-initiated SLO


arneroen:
I'm struggling to make IdP-initiated SLO work. As per several examples, I have the following routes:
```
app.post('/logout/callback', function(req, res) {
        req.logout();
        res.redirect('/');
});
app.get("/logout", function(req, res) {
        if (!req.user) res.redirect("/");
        console.log("initiating logout", req.log, req.user);
        return passport._strategy("saml").logout(req, function(err, uri) {
            return res.redirect(uri);
        });
});
```

In addition, logoutCallbackUrl is defined as '/logout/callback'. This works fine for SP-initiated logout. Obvivously not for IdP-initiated SLO, though, as being contacted at /logout/callback will not prompt the server to reply with a logoutResponse. I've not found any examples for how to implement IdP-initiated SLO. 

I've found [methods in the code](https://github.com/bergie/passport-saml/blob/master/lib/passport-saml/saml.js#L251), that are generating the logoutResponse, which leads me to believe that IdP-initiated SLO is probably implemented.  However, the beginning of the "chain of calls" to get to it is the "authenticate" functionality. 

I'd greatly appreciate any help with getting this to work! Thanks!

jonathanpdiaz:
hey @arneroen.
Have you found something else on this? I entered into the issues section to write exactly what you describe.
I'm getting the `LogoutRequest` from my Identity Provider (One Login), but the API of the strategy doesn't seem to handle it.
Any comment would be much appreciated.
Thanks!

Brynjulf:
@arneroen, @jonathanpdiaz  you should change your callback from a post to a get. That fixed the issue for me.

jonathanpdiaz:
I forgot to answer my own question after I found the issue. The library works perfectly.
In my case, we implemented the Single Logout using HTTP Redirect. 
What would this mean? It means that after I hit log out on any Service Provider [SP] (my app is one of them) passport-saml will generate a SAML Request XML that will log out this particular SP on the IdP and **the browser** will be redirected to the next SP that shares a session that needs to log out. To do this, the IdP will go to that particular SP and use the SLO endpoint.

One extra thing to consider is that the Identity Provider [IdP] will also hit the SLO endpoint to confirm the logout at the end. There, you can redirect your site to the login page.

Looks like this: (remember that is **the browser** who will be redirected to each site -one by one- that means that you have the information of the current session that is logged in each cookie that enables you to logout.)

[...]

PR #678, opened 2022-03-08
make RequestWithUser.samlLogoutRequest optional

clhuang:
# Description

whenever I try to do `passport.use('strategyName', new SamlStrategy(...))` I get the following error:
```
Argument of type 'import("/Users/calvinhuang/scale/scaleapi3/server/node_modules/passport-saml/lib/passport-saml/strategy").Strategy' is not assignable to parameter of type 'import("/Users/calvinhuang/scale/scaleapi3/server/node_modules/@types/passport/index").Strategy'
  Types of property 'authenticate' are incompatible.
    Type '(req: RequestWithUser, options: AuthenticateOptions) => void' is not assignable to type '(this: StrategyCreated<Strategy, Strategy & StrategyCreatedStatic>, req: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, options?: any) => any'.
      Types of parameters 'req' and 'req' are incompatible.
        Property 'samlLogoutRequest' is missing in type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>' but required in type 'RequestWithUser'.
```
It doesn't make sense to me that samlLogoutRequest needs to be present on every authentication request, and removing this would allow it to typecheck properly.

I'm using @types/passport 1.0.7.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -19,7 +19,7 @@ export interface StrategyOptions {
 export type User = Record<string, unknown>;
 
 export interface RequestWithUser extends express.Request {
-  samlLogoutRequest: Profile;
+  samlLogoutRequest?: Profile;
   user: User;
 }
 


srd90:
Commenting just this part of PR description:
>It doesn't make sense to me that samlLogoutRequest needs to be present on every authentication request, 

`passport-saml` has leaky/vulnerable SAML SLO implementation. 

It is used by propagating stuff from callback interfaces (authn and slo callback) to `authenticate`function, which is counter-intuitive but thats just the way it has been implemented back in the days. For further infromation see e.g. https://github.com/node-saml/passport-saml/issues/221#issuecomment-1048200363 and #419. 

There was some sort of "half-way" fix for it (see https://github.com/node-saml/node-saml/pull/10 and #619 ) which forces users of `passport-saml` library to take some/explicit responsibility of SLO scenarios and/or force users of `passpor-saml` library think of SLO instead of assuming that `passport-saml` would just work out of the box (which it doesn't...it just gave false sense of security and if someone used `passport-saml`'s SAML SLO implementation and did not bother to test it he/she could have made users of his/her service vulnerable to local session hijack).

cjbarth:
I would further like to note that `passport-saml` includes its own types, so you don't need to install other types.

clhuang:
Yeah, this is in regards to the `passport-saml` included types.

Is there something I'm missing in order to have the following code typecheck properly?
```ts
passport.use(new SamlStrategy(...))
```

clhuang:
this only seems to happen when there is a declaration for the `express.User` type somewhere in the project, e.g. custom.d.ts:

```ts
namespace Express {
  interface User {
    ...
  }
}
```

jbinto:
@clhuang Any chance you can share more about how you resolved this? We are just starting to see this error after upgrading from 3.2.1 to 4.01, and we also have modified express's `express.User` type.

jbinto:
Linking #549 to this issue.

Issue #606, opened 2021-06-05
[BUG] 3.0.0 typescript compiler: Could not find a declaration file for module 'passport-strategy'


rskvazh:
**To Reproduce**

1. `npm install passport passport-saml @types/passport`
2. `tsc`

Got error: 

```shell
node_modules/passport-saml/lib/passport-saml/strategy.d.ts:1:46
error TS7016: Could not find a declaration file for module 'passport-strategy'. '/Users/rskvazh/Projects/SpatialChat/spatialchat-universe/node_modules/passport-strategy/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/passport-strategy` if it exists or add a new declaration (.d.ts) file containing `declare module 'passport-strategy';`

1 import { Strategy as PassportStrategy } from "passport-strategy";
                                               ~~~~~~~~~~~~~~~~~~~
Found 1 error.
```

**Expected behavior**

`npm i --save-dev @types/passport-strategy` fixed the error and successfully compiled, but I think this should be fixed in a passport-saml package.

**Environment**

- Node.js version: v14.15.1
- passport-saml version: 3.0.0
- typescript: 4.3.2

markstos:
The problem here was that you didn't include `--dev` in your original install statement. We already have the correct version of `@types/passport-strategy` as a dev dependency. 

The module is not needed for "production" installs, so it's correct that we don't have it as a production dependency.

BigForNothing:
Can we reopen this? This is not the solution. @types/passport-strategy should be part of the dependencies, and not the devDependencies.

cjbarth:
Types should always be installed as dev dependencies. See https://github.com/DefinitelyTyped/DefinitelyTyped#npm

BigForNothing:
Their docs are (sort-of) wrong. See the official typescript website [documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies) for publishing:

> Our package exposes declarations from each of those, so any user of our browserify-typescript-extension package needs to have these dependencies as well. For that reason, we used "dependencies" and not "devDependencies", because otherwise our consumers would have needed to manually install those packages. If we had just written a command line application and not expected our package to be used as a library, we might have used devDependencies.

The "sort-of" is because it should go in to `devDependencies` if you aren't publishing (which is what your link. If you are and someone is consuming it (like passport-saml), it goes into dependencies.

For passport-saml:
```
"dependencies": {
[...]

PR #686, opened 2022-04-04
Move TypeScript-required types to dependencies from devDependencies

cjbarth:
# Description

Update use of types to match the recommendations of TypeScript.

# Checklist:

- Issue Addressed: [  #606 ]
- Link to spec: [ https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies ]
- Tests included? [ n/a ]
- Documentation updated? [ n/a ]

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -419,10 +419,9 @@
       "dev": true
     },
     "@types/body-parser": {
-      "version": "1.19.1",
-      "resolved": "",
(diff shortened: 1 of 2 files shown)


markstos:
At my company we also mostly gave up on using `devDependencies` and just use `dependencies`, so I'm not surprised this same kind of problem came up in the complex TypeScript ecosystem as well.

markstos (approved):
Reviewed diff.

PR #696, opened 2022-06-13
Update dependencies

cjbarth:
# Description

Simple update of dependencies such that all tests run after the update. There are other dependencies that need to be updated, but would involve code changes too.


markstos (approved):
Reviewed diff. Looks straightforward.

markstos:
Hmm, looks like the 14.x build causes an out of memory error. I presume this is caused by one of the new dependency updates. Maybe the responsible project has already figured out they are causing problems and have issued an even newer update?

PR #697, opened 2022-06-16
3.x

alesmenzel:
This is a breaking change that supports passport version `0.6.x` - see changelog https://github.com/jaredhanson/passport/blob/1e8f112bd233dbffb1904d4dd2051780d81b0a22/CHANGELOG.md#060---2022-05-20

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "watch": "concurrently --kill-others \"npm:*-watch\""
   },
   "dependencies": {
-    "@types/passport": "^1.0.9",
     "@xmldom/xmldom": "^0.7.5",
     "debug": "^4.3.2",
     "passport-strategy": "^1.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -60,8 +60,10 @@
   },
   "devDependencies": {
     "@types/debug": "^4.1.7",
+    "@types/express-session": "^1.17.4",
```
cjbarth, line 63: I believe all of these should actually be `dependencies` not `devDependencies`.

cjbarth (changes requested):
Thank you so much for your contribution! I noted this and was trying to make time to address it.

As I started reviewing this, I noticed that this isn't against `master`. Since this will be a breaking change, it should be against `master` where all the other 4.x changes are. Please make this against `master` before I continue reviewing it.

package.json
alesmenzel: Hi, the express-session is only used in tests. I will close this PR and reopen it on master.

alesmenzel:
Updating master instead -> https://github.com/node-saml/passport-saml/pull/698

PR #698, opened 2022-06-17
Update passport to 0.6.x

alesmenzel:
Update passport version to 0.6.x - see changelog https://github.com/jaredhanson/passport/blob/1e8f112bd233dbffb1904d4dd2051780d81b0a22/CHANGELOG.md#060---2022-05-20

This change should not break usage with passport 0.5.x since the callback used to be optional in that version.


cjbarth:
> This change should not break usage with passport 0.5.x since the callback was not needed in that version and just wont be used.

I agree, but the types will break. We actually should probably update the types to indicate that the callback is optional. Are you interested in doing that?

src/strategy.ts
```diff
@@ -104,7 +104,11 @@ export abstract class AbstractStrategy extends PassportStrategy {
             }
 
             // Log out the current user no matter if we can verify the logged in user === logout requested user
-            req.logout();
+            req.logout((err) => {
```
cjbarth, line 107: This really should block until the callback happens. The reason for this is that otherwise the code will complete running before `logout` actually is confirmed to happen or have errored. IIRC, your previous promise-based solution was better. For example, when you run the tests, you should see [this code](https://github.com/node-saml/passport-saml/blob/ef328daba4e9322346492c7cbfb3dbc1ccf2a13b/test/capturedSamlRequests.spec.ts#L929-L940) execute before the test finishes.

cjbarth (changes requested):

src/strategy.ts
alesmenzel: Updated the PR to wait for the logout.

alesmenzel: This change breaks the compatibility with 0.5 because it will now require req.logout(...) to call the promise executor callbacks, otherwise the promise will never settle.

cjbarth (approved):
I pushed some commits to reduce code duplication and make things easier to read. If you approve of these changes, please post a comment and then we'll get this landed.

PR #708, opened 2022-06-26
docs: move history note to the bottom and expand it.

markstos:
Before, one of the first sentences in the documentation was a history note about a fork that was made 10 years ago. That's not one of the first things people need to know about the project.

Now, a note on the history of the project has been moved to the bottom and expanded to credit the primary maintainers throughout the last decade.

During this commit, the project is about to turn 10 years old in 2 days! 

The project was born on June 28th, 2012 with a commit by @bergie.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@
 
 This is a [SAML 2.0](http://en.wikipedia.org/wiki/SAML_2.0) authentication provider for [Passport](http://passportjs.org/), the Node.js authentication library.
 
-The code was originally based on Michael Bosworth's [express-saml](https://github.com/bozzltron/express-saml) library.
-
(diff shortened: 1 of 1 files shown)


markstos:
@cjbarth Can 4.0 be released on June 28th? That's passport-saml's 10 year birthday!

cjbarth:
I will be unavailable to work on this project for the next few days, so I can't do a release on the anniversary. I know we're close, but there are lots of i's to dot and t's to cross before a release. Now that we actually have the beta versions out an on NPM, I think it will make it easier to get final feedback too. A lot has changed, and I don't want to mess people up with a security module that doesn't quite work.

markstos:
Sounds good, thanks!

PR #811, opened 2022-11-14
Simplified readme, fixes #809

### LINKED ISSUE #809: Readme is wrong
The readme shows the following example:

```
passport.use(
  new MultiSamlStrategy(
    {
      passReqToCallback: true, // makes req available in callback
      getSamlOptions: function (request, done) {
        findProvider(request, function (err, provider) {
          if (err) {
            return done(err);
          }
          return done(null, provider.configuration);
        });
      },
    },
    function (req, profile, done) {
      // for signon
      findByEmail(profile.email, function (err, user) {
        if (err) {
          return done(err);
        }
        return done(null, user);
      });
    },
    function (req, profile, done) {
      // for logout
      findByNameID(profile.nameID, function (err, user) {
        if (err) {
          return done(err);
        }
        return done(null, user);
      });
    }
  )
);
```

But this doesnt seem to be valid JavaScript. All three parameter contain a function within a function:

```
function (request, done) {
            findProvider(request, function (err, provider) {
```

Further the example shows `new SamlStrategy()` and `new MultiSamlStrategy()` with three arguments but according to the TypeScript interface the constructor only accepts two arguments. In this case I wonder if the Typings are wrong or the readme.

Bildschirm­foto 2022-11-10 um 16 11 50

MickL:
Simplified readme, fixes #809

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [X]

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,21 +41,19 @@ passport.use(
     },
     function (profile, done) {
       // for signon
-      findByEmail(profile.email, function (err, user) {
-        if (err) {
-          return done(err);
-        }
-        return done(null, user);
-      });
+      // ...
+      if (err) {
+        return done(err);
+      }
+      return done(null, user);
     },
     function (profile, done) {
       // for logout
-      findByNameID(profile.nameID, function (err, user) {
-        if (err) {
-          return done(err);
-        }
-        return done(null, user);
-      });
+      // ...
+      if (err) {
+        return done(err);
+      }
+      return done(null, user);
     }
   )
 );
@@ -84,21 +82,19 @@ passport.use(
     },
     function (req, profile, done) {
       // for signon
-      findByEmail(profile.email, function (err, user) {
-        if (err) {
-          return done(err);
-        }
-        return done(null, user);
-      });
+      // ...
+      if (err) {
+        return done(err);
+      }
+      return done(null, user);
     },
     function (req, profile, done) {
       // for logout
-      findByNameID(profile.nameID, function (err, user) {
-        if (err) {
-          return done(err);
-        }
-        return done(null, user);
-      });
+      // ...
+      if (err) {
+        return done(err);
+      }
+      return done(null, user);
     }
   )
 );


cjbarth:
Thank you for this PR, but I think it is missing something important. What this README is trying to capture is that you have some custom functions i.e. `findByNameId` and `findByEmail`, which are your custom functions that are called when those signin and logout functions are called.

MickL:
> What this README is trying to capture is that you have some custom functions

This is what I tried with the PR: Leave it open so people know this is the point where they implement THEIR logic. Using this functions `findByNameId` really confused me, especially as the second argument is another function. Personally I would never pass a function into a function.

cjbarth:
> Personally I would never pass a function into a function.

You probably do all the time, it is the callback. If you mean `findByNameId()`, that function can live anywhere at the top scope of your module/code, just like any other function you might call, like `console.log()`.

MickL:
Passing a function into findByNameId() as the second argument is an antipattern IMO. It was super confusing for me as I didnt see that for ages. It creates a callback hell and also the function findByNameId() is not capsulated testable this way and hard to debug.

If you want to display a function then I would at least dont pass in another function into it and instead await it. 

```
async function (profile, done) {
       // for signon
       try {
           const user = await findByEmail(profile.email);
           return done(null, err);
       } catch (err) {
           return done(err);
       }
 }
```
          
But instead my suggestion is to not show any implementation example at all and just show `// ...`

cjbarth:
You can't use `async` and `await` with Passport, you have to use callbacks. Callbacks aren't an antipattern at all. Have a look at this [introduction to NodeJS](https://nodejs.org/en/docs/guides/getting-started-guide/). You'll see that the last argument to `server.listen()` is a function, which is the callback which runs after the server has started listening. This is a fundamental part of asynchronous programming and predates `async/await` syntax. In fact, since `async/await` is based on `Promises`, and [`Promises` fundamentally work on callbacks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) (you have to pass two callbacks to `new` up a `Promise`, understanding callbacks is key to understanding how `async/await` work behind the scenes.

PR #838, opened 2023-01-12
StrategyOptionsCallback shouldn't have to pass all SAML options

oliverlockwood:
…since some can be derived from the defaults set up during initialisation of the strategy, as per the documentation at http://www.passportjs.org/packages/passport-saml/, which clearly says:

> The options passed when the MultiSamlStrategy is initialized are also passed as default values to each provider. e.g. If you provide an issuer on MultiSamlStrategy, this will be also a default value for every provider. You can override these defaults by passing a new value through the getSamlOptions function.

This fixes https://github.com/node-saml/node-saml/issues/246.
(I raised the issue in a different repo, as that's where the `SamlConfig` class resides, but it turns out the change I thought to propose is in this repo. Sorry.)


cjbarth:
Can we get some tests to make sure this behavior doesn't break in the future? Ideally a tests that throws a type error (which you should mark as such with [@ts-expect-error](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) and then a test that works as you want it to, which is currently broken.

cjbarth:
@oliverlockwood , I'm preparing a release soon. Would you like to work with me to get this landed for the next release?

cjbarth (changes requested):
Needs tests.

oliverlockwood:
Hi @cjbarth, yes, I'll give this a crack.  Thanks for the reminder.

test/multiSamlStrategy.spec.ts
```diff
@@ -25,6 +25,18 @@ describe("MultiSamlStrategy()", function () {
     expect(strategy).to.be.an.instanceOf(Strategy);
   });
 
+  it("does not require issuer in the SamlOptionsCallback", function () {
```
cjbarth, line 28: This is an inadequate test. It simply defines a function; it never uses it. With this test we have no way of knowing if this function is fit for purposes.

cjbarth (changes requested):
This needs a test that actually exercises the fact that `issuer` isn't required in `SamlOptionsCallback`.

oliverlockwood:
With the "stale" reminder, I'll be honest - I haven't had time to write more extensive tests for this, and it doesn't feel like that's going to change any time soon.  I'm sorry about that, but I'm just being realistic.

@cjbarth I guess you'll have to just make a decision as to whether to accept this trivial correction without UT coverage being provided, or whether you'll reject it and continue to have a mismatch between documentation and type definitions.  Your call.

cjbarth (approved):

PR #854, opened 2023-04-05
Allow to configure optional logger function to log SAML operations

RajuTanchak:
# Description

Extended `StrategyOptions` to define optional logger function to log SAML request/response with respect to node-saml library [issue](https://github.com/node-saml/node-saml/pull/262).

# Checklist:

- Issue Addressed: [#279]
- Link to SAML spec: [ ]
- Tests included? [No]
- Documentation updated? [Yes]


RajuTanchak:
> What kind of information are you looking for when logging? Our logging of problems has gotten pretty good since the related issue was opened.

Need to supply `log` function in `StrategyOptions` with respect to changes done on [this](https://github.com/node-saml/node-saml/pull/262) PR.

cjbarth:
It seems that this isn't going to more forward. It can be reopened if it does.

PR #855, opened 2023-04-11
Use secure version of node-saml

cjbarth:
# Description

node-saml has been updated with a patch release to address a security concern. Set the minimum allowed version of node-saml to make sure we get that security patch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -564,20 +564,20 @@
       }
     },
     "@node-saml/node-saml": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/@node-saml/node-saml/-/node-saml-4.0.0.tgz",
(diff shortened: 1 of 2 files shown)


markstos:
For those looking for the vuln details:

* [xml2js had a prototype pollution vuln](https://github.com/Leonidas-from-XIV/node-xml2js/issues/664)
* [xml-encryption had an insecure dependency in `@xmldom/xmldom`, which allows multiple root elements in a DOM tree](https://github.com/xmldom/xmldom/security/advisories/GHSA-crh6-fp67-6883)

I'm sure if these vulns are accessible through passport-saml or not, but considering that we pass through XML to be parsed by these libraries, it's best to presume that they are reachable through passport-saml.

PR #856, opened 2023-04-18
Return unauthorized if no samlOptions is provided

lanker:
If a saml config can't be found for a user in a `multiSamlStrategy` setup, it probably means that the user is not authorized to login (via saml), so return 401 in those cases. Previously, missing `samlOptions` raised an exception in node-saml where the check for `samlOptions.issuer` failed.

# Checklist:

- Issue Addressed: [-]
- Link to SAML spec: [-]
- Tests included? [x]
- Documentation updated? [-] (couldn't find any documentation for the current behavior)

### DIFF
--- a/src/multiSamlStrategy.ts
+++ b/src/multiSamlStrategy.ts
@@ -45,6 +45,10 @@ export class MultiSamlStrategy extends AbstractStrategy {
         return this.error(err);
       }
 
+      if (!samlOptions) {
+        return this.fail(401);
+      }
+
       const samlService = new SAML({ ...this._options, ...samlOptions });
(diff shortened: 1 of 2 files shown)


cjbarth:
Why would you not want to handle the exception in your own code and return an appropriate response? I'm not sure every consumer of this library would want the same behavior if `samlOptions` is missing.

lanker:
Yeah, I actually didn't realize it was possible until after doing this PR, but kept it open to hear your opinion. I think after figuring out how it can be handled I'm OK with the current behavior, so I'll close this PR. Thank you for your time and your work on this!

PR #858, opened 2023-04-19
docs(README): How to migrate from legacy passport-saml

MrSwitch:
Thankyou for the library 🙏

I, and a few people (judging by some issues) have had troubles migrating from the legacy [`passport-saml`](https://www.npmjs.com/package/passport-saml) library, which recently triggered an an unfixable `npm audit`... and is how i arrived here 😀.

For your consideration, i've propose a draft update to the docs, which i would hope might help some people.

# Description

This updates the documentation to:

- address the migration from the original `passport-saml` library. 
- includes a missing undocumented option crucial to backwards compatibility.

# Checklist:

- Documentation updated? [x]

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,7 +410,7 @@ passport.use(
     {
        // ...
        wantAuthnResponseSigned: false,
-			 wantAssertionsSigned: false,
+       wantAssertionsSigned: false,
     },
     ...
 ```
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ Please see the [type specification](https://github.com/node-saml/node-saml/blob/
 - `additionalAuthorizeParams`: dictionary of additional query params to add to 'authorize' requests
 - `identifierFormat`: optional name identifier format to request from identity provider (default: `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`)
 - `wantAssertionsSigned`: if truthy, add `WantAssertionsSigned="true"` to the metadata, to specify that the IdP should always sign the assertions.
+- `wantAuthnResponseSigned` (default: true): if true, require that all incoming authentication response messages be signed at the top level, not just at the assertions. It is on by default. Note: either the response or the assertion must be signed even if both are turned off.
 - `acceptedClockSkewMs`: Time in milliseconds of skew that is acceptable between client and server when checking `NotBefore` and `NotOnOrAfter` assertion condition validity timestamps. Setting to `-1` will disable checking these conditions entirely. Default is `0`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,6 +138,7 @@ Please see the [type specification](https://github.com/node-saml/node-saml/blob/
 - `additionalAuthorizeParams`: dictionary of additional query params to add to 'authorize' requests
 - `identifierFormat`: optional name identifier format to request from identity provider (default: `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`)
 - `wantAssertionsSigned`: if truthy, add `WantAssertionsSigned="true"` to the metadata, to specify that the IdP should always sign the assertions.
+- `wantAuthnResponseSigned` (default: true): if true, require that all incoming authentication response messages be signed at the top level, not just at the assertions. It is on by default. Note: either the response or the assertion must be signed even if both are turned off.
```
cjbarth, line 141: This is good, but the note at then end should probably be on the other note for `wantAssertionsSigned` as well.

README.md
```diff
@@ -399,6 +400,21 @@ See [Releases](https://github.com/node-saml/passport-saml/releases) to find the
 
 ## FAQ
 
+### How do i migrate from `passport-saml`
+
+The original `passport-saml` is now deprecated. To migrate to this version of `passport-saml` you'll need to use the following options which have a new default behaviour.
```
cjbarth, line 405: Doesn't this advice contradict your note above?

At any rate, we don't want to suggest turning off all security because we don't want any copy-paste security issues for those that use this library. You can describe the problem, but I would advise againt code that can be copied.

cjbarth (changes requested):
I appreciate the need for this, but would a migration document be better suited for a Wiki page? Also, the configuration options in this README should only be for this project. Ones from `node-saml` should only be referenced in this README.

README.md
MrSwitch: Thanks @srd90.

@cjbarth i'm quite happy with @srd90 suggestion. How do you feel about doing that?

cjbarth: I'd like to hear @markstos  thoughts, but we've been trying to remove properties from this README that are actually related to `node-saml` so that we don't have to maintain them in two places. I'd actually like to see continued effort in that direction with a note to indicate that the README at `node-saml` should be consulted.

cjbarth:
Closing because of inactivity and the general direction taken here is to make things less secure. This seems like it would be good wiki material

MrSwitch:
Sorry slipped my mind. Thanks for considering it.

PR #861, opened 2023-05-23
Revised AbstractStrategy for authenticate method to match PassportStrategy expectation

algoritmodourado:
Revised AbstractStrategy for authenticate method to match PassportStrategy expectation

# Description

The previous implementation was using the RequestWithUser internal type that contradict the base PassportStrategy authenticate given that user property SHOULD NOT be required for this process.
This causes a typings issue with the base passport which overrides the express Request interface to have a user property that is of type User | undefined;

# Checklist:

- Issue Addressed: N/A
- Link to SAML spec: N/A
- Tests included? No, fixing a typings issue with core base package
- Documentation updated? Not needed, seamless change

### DIFF
--- a/src/multiSamlStrategy.ts
+++ b/src/multiSamlStrategy.ts
@@ -39,7 +39,7 @@ export class MultiSamlStrategy extends AbstractStrategy {
     this._options = samlConfig;
   }
 
-  authenticate(req: RequestWithUser, options: AuthenticateOptions): void {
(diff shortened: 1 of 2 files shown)


cjbarth (changes requested):
Please remove updates to packages from this PR. We can do that in another PR.

algoritmodourado:
Removed package updates and rebased against current master

cjbarth (approved):

PR #864, opened 2023-06-02
Add bot to close stale issues

cjbarth:
Add a bot to close issues after 120 days being open with no activity. This is a long time for sure, but it does help with some house-keeping. We can adjust the numbers as needed, but this gives us a starting point on this sometimes-slow-moving project.


markstos:
I agree this is useful and we can tune the value later.

Issue #839, opened 2023-01-17
After upgrade from passport-saml to @node-saml/passport-saml logins stopped to working with an error SAML.validatePostResponseAsync, Invalid document signature


v0idmaster:
Hello, I upgraded from passport-saml to @node-saml/passport-saml and my logins stopped working in assertion validation.
I am using HTTP-POST binsings and a PingFederate IDP. The error I am getting is: 
Error: Invalid document signature
    at SAML.validatePostResponseAsync (C:\Develop\test\node_modules\@node-saml\node-saml\lib\saml.js:510:23)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
After debugging the code, I found out that it fails because it searches for a signature element having Reference descendant with an URI attr that is equal to the ID of the Result node, which is not the case and fails because it cannot find a signatue.

  const xpathTransformQuery =
    ".//*[" +
    "local-name(.)='Transform' and " +
    "namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#' and " +
    "ancestor::*[local-name(.)='Reference' and @URI='#" +
    currentNode.getAttribute("ID") +
    "']" +
    "]";

Here, currentNode is the Result element passed from the validatePostResponseAsync function (doc.documentElement). However, the Reference URL attr in my case ia actually equal to the ID of the Assertion element.

<samlp:Response Version="2.0" ID="ZHO2f87e-GRdaDjH0XdJJGD3pHT" IssueInstant="2023-01-16T20:28:51.587Z"
    xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol">
    <saml:Issuer
        xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">https://solvelocalassertion2
    </saml:Issuer>
    <samlp:Status>
        <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
    </samlp:Status>
    <saml:Assertion ID="w82kePprRVPi.hkkZ5QlwHD_DW_" IssueInstant="2023-01-16T20:28:51.587Z" Version="2.0"
        xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">
        <saml:Issuer>https://solvelocalassertion2</saml:Issuer>
        <ds:Signature
            xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
            <ds:SignedInfo>
[...]

PR #867, opened 2023-06-08
Roll-up changelog entries for beta releases

cjbarth:
@markstos , this is a PR that addresses your concerns mentioned in https://github.com/node-saml/passport-saml/issues/839#issuecomment-1582933682. If we want any of the lines in this changelog altered, we just have to change the name of the PR and regenerate the changelog.

I can do this same thing with `node-saml` if we like it.

### DIFF
--- a/.grenrc.js
+++ b/.grenrc.js
@@ -37,21 +37,17 @@ module.exports = {
         placeholders.name,
         `[${placeholders.text}](${placeholders.url})`,
       ];
-      return parts.filter((_) => _).join(" ");
+      return parts
+        .filter((_) => _)
+        .join(" ")
+        .replace("  ", " ");
     },
     release: function (placeholders) {
-      let dateParts = placeholders.date.split("/");
-      let placeholdersDate = new Date(
-        Number(dateParts[2]),
-        Number(dateParts[1]) - 1,
-        Number(dateParts[0])
-      );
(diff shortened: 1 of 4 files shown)


markstos (approved):
Thanks, Chris. Maybe during a beta cycle we want to break out the specific changes between betas, but now it does seem more useful to have all the 4.x changes bunched together. 

I updated a couple of PR titles for upgrading passport and node-saml to also add "see changelog for breaking changes". 

I had just tried to upgrade to passport 4.x myself and also got stung by breaking changes I missed until production. 🤦‍♂️ The challenge was the breaking changes were spread between passport and node-saml as well. And from reading the passport-saml change, the language of simply "update dependency" didn't convey the risk, because minor dependency updates happen all the change.

So I hope by adding PR titles like "update passport -- see changelog for breaking changes", that will help diligent upgraders find all the details they need for a smooth upgrade. 

I'll review the node-saml changelog now.

cjbarth:
No worries. Once you have `passport-saml` adjusted to your liking, I'll update this PR and then go over to `node-saml` and do the same thing. We can iterate until you're happy. One of these days I should see if we can get a GitHub action to do that work...

cjbarth:
I've made some updates @markstos . What do you think now?

markstos (approved):
Looks good! 
image

PR #880, opened 2023-09-08
Fix circular references #879

### LINKED ISSUE #879: Importing passport-saml fails on node v20.6.0
**Summary**

We've just encountered the issue that _importing_ (not running, but importing) passport-saml throws an error when:
- Node 20.6.0 is used
- you're importing passport-saml into an ESM module

Importing passport-saml will throw this error:

```
class MultiSamlStrategy extends strategy_1.AbstractStrategy {
                                           ^

TypeError: Class extends value undefined is not a constructor or null
    at Object.<anonymous> (~/saml/node_modules/@node-saml/passport-saml/lib/multiSamlStrategy.js:6:44)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Module._load (node:internal/modules/cjs/loader:938:12)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at require (node:internal/modules/helpers:130:18)
    at Object.<anonymous> (~/saml/node_modules/@node-saml/passport-saml/lib/index.js:21:29)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)

Node.js v20.6.0
```

Using an older node version works fine, and importing passport-saml into a CJS module works fine as well.

**To Reproduce**

See here for an MWE: https://github.com/LaurensRietveld/saml-mwe and follow the readme:

- Make sure you're running node 20.6.0
- Run `yarn`
- Run `node ./test.cjs`. This works fine
- Run `node ./test.js`. This throw the following error:

**Probable cause**

I expect this be caused by circular reference issue (given that this results in undefined imports). I noticed two circular references:
- `MultiSamlStrategy.js` [depends on](https://github.com/node-saml/passport-saml/blob/master/src/multiSamlStrategy.ts#L11) on `index.js`, which [depends on](https://github.com/node-saml/passport-saml/blob/master/src/index.ts#L2) `MultiSamlStrategy.js`
- `MultiSamlStrategy.js` [depends on](https://github.com/node-saml/passport-saml/blob/master/src/multiSamlStrategy.ts#L1) `strategy.js`, which [depends on](https://github.com/node-saml/passport-saml/blob/master/src/strategy.ts#L4) `index.js` which [depends on](https://github.com/node-saml/passport-saml/blob/master/src/index.ts#L2) `MultiSamlStrategy.js`

**Environment**

- Node.js version: 20.6.0
- `passport-saml` version: 4.0.4

LaurensRietveld:
# Description

This fixes #879 . Fixed this by breaking the circular chain, ensuring that we don't import the `./index.js` file in other files.
I did not include tests, as this depends on the node version, and whether the package runs in ESM mode

# Checklist:

- Issue Addressed: #879
- Link to SAML spec: N/A
- Tests included? No
- Documentation updated? No


src/strategy.ts
```diff
@@ -1,7 +1,8 @@
 import { Strategy as PassportStrategy } from "passport-strategy";
 import { strict as assert } from "assert";
 import * as url from "url";
-import { Profile, SAML } from ".";
+import { Profile } from ".";
```
LaurensRietveld, line 4: note that this import is fine: it's a type import, and will be omitted from the transpiled multisamlstrategy.js file

LaurensRietveld:
not stale, this is still an issue, and this PR still fixes it

cjbarth:
The bot is just doing its job. We haven't forgotten about this. We're trying to work through getting `node-saml` released with the updated version of `xml-crypto` before we tackle this. Thank you for pinging this to keep it alive.

kamronbatman:
xml-crypto v6 comes out, does that have the fix? (I am not sure what fix you are waiting for, so I am sorry for the ignorance).

cjbarth:
Yea, that contains part of the fix. I'm working on getting it into `node-saml`. Once that comes out, then we're on to this one. Should be soon; just waiting on some PRs to be signed off on.

cjbarth:
@LaurensRietveld , this PR is the next in line to get landed. Can you think of anyway to test this to make sure that the problem doesn't re-appear?

LaurensRietveld:
> @LaurensRietveld , this PR is the next in line to get landed. Can you think of anyway to test this to make sure that the problem doesn't re-appear?

Good to hear!

I thought about testing this properly, but think that's intricate: my problem only occurs when 1) there is a circular dependency, and 2) when importing node-saml from an ESM module.
Automating this is tricky.

cjbarth (approved):

Issue #859, opened 2023-04-20
[BUG] Invalid document signature


royalcala:
Could you help me to know why I have this error after we updated from passport-saml to @node-saml/passport-saml:

Error: Invalid document signature
    at SAML.validatePostResponseAsync (/usr/src/server/node_modules/@node-saml/node-saml/lib/saml.js:510:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)

cjbarth:
Perhaps your issue is the same as https://github.com/node-saml/passport-saml/issues/839 or https://github.com/node-saml/passport-saml/issues/816? If not, please post more information about the versions your running and a minimal test case. Ideally we'd like to see a PR with a test that fails.

catamphetamine:
We also encountered such error and the workaround was adding the following options:

```js
    // After updating from `passport-saml` to `@node-saml/passport-saml`,
    // it started throwing `Error: Invalid document signature`
    // when testing SSO using Acadeum's Google SAML server.
    // Turns out that's a common issue and the workaround is to pass
    // `wantAssertionsSigned: false` / `wantAuthnResponseSigned: false` flags.
    // https://github.com/node-saml/passport-saml/issues/839
    // https://github.com/node-saml/node-saml/blob/master/CHANGELOG.md#v400-2022-10-28
    wantAssertionsSigned: false,
    wantAuthnResponseSigned: false,
```

edwandr:
I encountered the same issue and after quite some digging in code this seems to be indeed because in `validatePostResponseAsync` method, the default value in `this.options.wantAssertionsSigned` is true if not specified in SamlConfig whereas it was false in v3.

@cjbarth are you sure this is an intended behavior ? I feel `wantAssertionsSigned` should be false if not specified in config

cjbarth:
The change was intended as it is in harmony with the secure-by-default philosophy that we have here. Please feel free to put up a PR to help us with our README.

edwandr:
Ok I understand but then indeed it would be great to specify it in documentation, I can open a PR for that 👍
What would have been really helpful also is to add it as a breaking change in the changelog, I don't know if it's automatically generated or if it's possible to also update it now

markstos:
I agree this would be useful to highlight as a breaking change in the Changelog.

edwandr:
I opened a PR to update both doc and changelogs: https://github.com/node-saml/passport-saml/pull/883

PR #883, opened 2023-11-28
Specify wantAssertionsSigned default in CHANGELOG & README

edwandr:
# Description

An update of documentation and changelog to specify that `wantAssertionsSigned` configuration option is now `true` by default and must be declared explicitly as `false` if the IdP does not sign assertions.

This is a breaking change when switching from v3 to v4 on a project with unsigned assertions that does not declare `wantAssertionsSigned`, cf this issue: https://github.com/node-saml/passport-saml/issues/859
 
# Checklist:

- Issue Addressed: https://github.com/node-saml/passport-saml/issues/859
- Link to SAML spec: N/A
- Tests included? N/A
- Documentation updated? Yes


markstos:
@cjbarth Is there a pattern to follow when we want to make exceptions to the auto-generated Changelog content?

CHANGELOG.md
```diff
@@ -63,6 +63,7 @@
 - Update packages; set minimum Node to 14 [#685](https://github.com/node-saml/passport-saml/pull/685)
 - [**security**] Check user matches logout request before reporting logout success [#619](https://github.com/node-saml/passport-saml/pull/619)
 - Remove `node-saml` code and use an import instead [#612](https://github.com/node-saml/passport-saml/pull/612)
+- Changed `wantAssertionsSigned` default value to `true`. If your IdP does not sign assertions, you should specify it explicitly to `false` in strategy configuration
```
cjbarth, line 66: Do you know what PR caused this?

cjbarth: I would just mark that PR with a label of `breaking-change` and then make sure the description is accurate. The thing is, we did do a semver-major bump with that, so it really should be assumed there would be breaking changes, and it is typically up to the consumer to check transient dependencies.

I think what would really help in this case would be to clean up this README to explicitly point to the node-saml README for information. In this way, it really forces people to check the correct source for documentation and the documentation won't drift.

cjbarth: I'm in complete agreement with you on your two perspectives. The question is: how to proceed? I don't have the bandwidth to do more than I am. We can't even get anyone from the community to update the existing broken docs.

cjbarth: We'd love to have someone help with documentation updates. Ideally, we'd have a link to the documentation over at `node-saml` for all those properties so that there isn't any duplication. Feel free to update this PR or make others PRs to advance this goal. Thank you for being willing to help with this.

edwandr: I'm sorry but sadly I won't have the time to handle that kind of work in the coming months :/ I'll close this PR as it doesn't answer well to the problem

PR #952, opened 2025-02-05
express 5.x. move types to devDependencies

charlie-harvey:
# Description

Express 5 is out! NestJS 11 is out! It relies on Express 5. But I can't do cool saml things because this library pushes `@types/express` on me at version `4`. So I updated the package.json. That's all.

# Checklist:

- Issue Addressed: [Express 5 ]

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,15 @@
 {
   "name": "@node-saml/passport-saml",
-  "version": "5.0.0",
+  "version": "6.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@node-saml/passport-saml",
-      "version": "5.0.0",
+      "version": "6.0.0",
       "license": "MIT",
       "dependencies": {
         "@node-saml/node-saml": "^5.0.0",
-        "@types/express": "^4.17.21",
-        "@types/passport": "^1.0.16",
-        "@types/passport-strategy": "^0.2.38",
         "passport": "^0.7.0",
         "passport-strategy": "^1.0.0"
       },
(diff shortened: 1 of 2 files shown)


shlomisas:
Any update on this one?? it also holds up back merge an important PR

markstos (approved):
I'm aligned with this.

It should not be the job of this project to ship the types for Express, Passport or Passport-Strategy.

Also, fine to use Express 5 instead of 4 for testing.

cjbarth:
This is not an appropriate change because we export types that inherit from other types, so we have to export them for this all to work. As an example see https://github.com/node-saml/passport-saml/blob/113412a096486823d010f1f494df1bdf87f9cbf8/src/strategy.ts#L16.

cjbarth:
We're also not going to take on a semver major bump for a single update. We'll do it with the deprecation of a Node version and other updates to dependencies all around.

cjbarth:
We are preparing a round of major releases starting with the xml-crypto library. We'll roll that up to node-saml with its semver major release, then we'll do the same here with passport-saml. If this is really blocking you, I suggest temporarily using a fork until this work is completed.

Xenope:
Any news on this one, would love to upgrade with express5 compatibility.

KiwiKilian:
Until this is published you can use this in your `package.json`:

```json
  "overrides": {
    "@node-saml/passport-saml": {
      "@types/express": "5.0.3"
    }
  }
```

KiwiKilian:
@cjbarth because you closed this, I hope this is still on the agenda for another major release? Maybe a decoupling of the Express version is also possible, so users can provide their own version/request?

cjbarth:
Yes, we will update dependencies like this with the next breaking release. We avoid semver major dependency updates without taking our own semver major bump.

fxsalazar:
@cjbarth any news on this?

cjbarth:
I'm a little behind on some work with this organization. I haven't forgotten about it though. I'm focusing on updating the dependencies of this project first.

